PR #30, opened 2016-12-30
WIP: RebootGroup TPR

mischief:


dghubble:
TPR's are being replaced with Custom Resource Definitions in Kubernetes 1.7. We have an effort to migrate all TPR's in use to CRD's. This PR would require a redesign so I'm closing for now.

PR #32, opened 2016-12-30
hack: wip integration test

mischief:


dghubble:
This has useful work we might be able to borrow in future, but can't merge without some effort. Integration tests aren't on the immediate roadmap though...

Issue #61, opened 2017-04-20
Should not clobber new node fields (like taints)


euank:
Right now the components use Update, not Patch.

See https://github.com/coreos/flannel/issues/667 for why this is a problem.

We need to switch to patch, ideally sooner rather than later.

euank:
Container Linux is deprecated. Closing this issue as stale.

PR #62, opened 2017-04-24
Add agent daemonset management, support 1.6 & master nodes, some cleanup

### LINKED ISSUE #57: Operator: manage the lifecycle of the associated agent daemonset
This allows the update story to be nicer and allows the operator to ensure the right version is deployed, as well as configured correctly.

This should tie-in with #30 since it makes sense for this image to be configurable via a TPR option.

euank:
Lots of changes, all only manually tested.

Basically, this is what I think is the minimal set of changes needed to go into the tectonic-installer repo.

Notably, the update-management is simplified via the operator taking some of that responsibility, and the remaining code is updated for 1.6.

This fixes #57, and makes it so #61 won't matter again until 1.7 (we need to switch to patch before then though).

There is a k8s vendor bump in here, so beware the diff viewer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Currently, `update-operator` only reboots one node at a time.
 
 ## Requirements
 
-- Working Kubernetes >= 1.4 on CoreOS
+- Working Kubernetes >= 1.6 on CoreOS
 - `update-engine.service` should be unmasked, enabled and started in systemd
 - `locksmithd.service` should be masked and stopped in systemd
 


pkg/operator/agent_manager.go
```diff
@@ -0,0 +1,14 @@
+package operator
+
+import "k8s.io/client-go/pkg/api/v1"
+
+var managedByOperatorLabel = map[string]string{
+	"managed-by": "prometheus-operator",
```
crawford: CLUO

pkg/constants/constants.go
```diff
@@ -61,4 +61,10 @@ const (
 
 	// Key set by the update-agent to the value of "VERSION" in /etc/os-release.
 	LabelVersion = Prefix + "version"
+
+	// AgentVersion is the key used to indicate the
+	// container-linux-update-operator's agent's version.
+	// The value is a semver-parseable string. It should be present on each agent
+	// pod, as well as on the daemonset that manages them.
+	AgentVersion = Prefix + "agent-versoin"
```
crawford: version

Makefile
```diff
@@ -5,7 +5,9 @@ all:	bin
 bin:	bin/update-agent bin/update-operator
 
 bin/%:
-	CGO_ENABLED=0 go build -ldflags '-s -w' -tags netgo -o $@ ./cmd/$*
+	CGO_ENABLED=0 go build \
+							-ldflags '-s -w -X github.com/coreos/container-linux-update-operator/pkg/version.Version=$(shell cat VERSION) -X github.com/coreos/container-linux-update-operator/pkg/version.Commit=$(shell git rev-parse HEAD)' \
```
crawford: Do these need to be spaces after the first tab?

VERSION
```diff
@@ -0,0 +1 @@
+0.2.0
```
crawford: Why not use git tags?

euank: I think it's easier to reason about having the file there, especially since it's needed for every build (including non-tagged builds). Getting the most recent tag and then updating it depending on whether it's a dev or release build seems kinda messy.

The plan is to add a 'tag-release' script which sets the appropriate version in that file, git tags and commits, and then bumps the minor version again for the current "dev" version.

There's room for improvement, but I don't see a problem with a VERSION file.

Makefile
euank: probably, honestly this should be in a `./build/` script probably.

euank:
This PR got broken, github lost track of the ref so I couldn't update it anymore.
Opening a new one...

PR #63, opened 2017-04-24
Add agent daemonset management, support 1.6 & master nodes, some cleanup

### LINKED ISSUE #57: Operator: manage the lifecycle of the associated agent daemonset

euank:
Lots of changes, all only manually tested.

Basically, this is what I think is the minimal set of changes needed to go into the tectonic-installer repo.

Notably, the update-management is simplified via the operator taking some of that responsibility, and the remaining code is updated for 1.6.

This fixes #57, and makes it so #61 won't matter again until 1.7 (we need to switch to patch before then though).

There is a k8s vendor bump in here, so beware the diff viewer.


euank:
See #62 for a short bit of existing discussion. This is a re-opening of it due to a github bug.

pkg/operator/agent_manager.go
```diff
@@ -1,14 +1,221 @@
...
+		}
+		dsSemver = ver
+	} else {
+		glog.Errorf("managed daemonset did not have a version annotation: %+v", agentDS)
+		return fmt.Errorf("managed daemonset did not have a version annotation")
+	}
+
+	if dsSemver.LT(version.Semver) {
+		// daemonset is too old, update it
+		// TODO: perform a proper rolling update rather than delete-then-recreate
```
derekparker: Once https://github.com/coreos-inc/operator-client/pull/41 lands we can use that to do a proper upstream rolling update. Even without that patch, older versions of operator-client working on a < 1.6 cluster will do a manual rolling update.

pkg/operator/agent_manager.go
```diff
@@ -1,14 +1,221 @@
...
+		// Right now, daemonset rolling updates aren't upstream and are thus fairly
+		// painful to do correctly. In addition, doing it correctly doesn't add too
+		// much value unless we have corresponding detection/rollback logic.
+		falseVal := false
+		err := k.kc.DaemonSets(k.namespace).Delete(agentDS.Name, &v1.DeleteOptions{
+			OrphanDependents: &falseVal, // Cascading delete
+		})
+		// TODO: this requires delete to be blocking to work. Is it?
```
derekparker: I think `Delete` is blocking, but have you verified?

euank:
@derekparker I was going to use operator-client, but since it's not open-source I can't reference it here, and also it doesn't have the same version / label selector semantics I'm using.

I do think the version-labels make this more restart-safe and so are a good idea.

pkg/operator/agent_manager.go
euank: In practice it always succeeds, so good enough for me to delete that comment.

The failure case is it restarting and trying again, so I'll just remove the comment.

pkg/operator/operator.go
```diff
@@ -114,18 +118,17 @@ func New() (*Kontroller, error) {
 		Component: leaderElectionEventSourceComponent,
 	})
 
-	return &Kontroller{kc, nc, er, leaderElectionClient, leaderElectionEventRecorder}, nil
+	namespace := os.Getenv("POD_NAMESPACE")
+	if namespace == "" {
+		return nil, fmt.Errorf("unable to determine operator namespace: please ensure POD_NAMESPACE environment variable is set")
```
euank: I'd rather leave as is for consistency. *shrug.

pkg/operator/operator.go
```diff
@@ -214,6 +214,19 @@ func (k *Kontroller) Run(ctx context.Context) error {
...
 
+		if manageAgent {
+			agentsReady, err := xyz()
+			if err != nil {
+				glog.Errorf("unable to ensure managed agents are ready: %v", err)
+				continue
+			}
+			if !agentsReady {
+				glog.V(4).Infof("waiting for managed agents to be ready, not doing additional work now")
+				// TODO: if this has happened for a long enough duration, it should become an error.
```
euank: This code was changed in a later version to poll via crashlooping if necessary (the agentsReady variable was entirely removed)

pkg/operator/operator.go
```diff
@@ -175,33 +174,23 @@ func (k *Kontroller) withLeaderElection(ctx context.Context) (context.Context, e
 					waitLeading <- struct{}{}
 				},
 				OnStoppedLeading: func() {
-					glog.V(5).Info("stopped leading; calling cancel")
-					cancelFunc()
+					glog.Fatalf("leaderelection lost")
```
euank: I'd rather leave it if you don't feel strongly.

pkg/operator/agent_manager.go
```diff
@@ -0,0 +1,14 @@
+package operator
+
+import "k8s.io/client-go/pkg/api/v1"
+
+var managedByOperatorLabel = map[string]string{
+	"managed-by": "container-linux-update-operator",
+}
+
+// updateAgent updates the agent on nodes if necessary.
+// It returns a value indicating whethre the update-agent is currently up to
```
euank: Comment already deleted

pkg/operator/agent_manager.go
```diff
@@ -1,14 +1,221 @@
...
+	// Each agent daemonset includes the version of the agent in the selector.
+	// This ensures that the 'orphan adoption' logic doesn't kick in for these
+	// daemonsets.
+	versionedSelector := make(map[string]string)
+	for k, v := range managedByOperatorLabels {
+		versionedSelector[k] = v
+	}
+	versionedSelector[constants.AgentVersion] = version.Version
+
+	return &v1beta1.DaemonSet{
```
yifan-gu: nit: Embed as string and unmarshal them would be easier to read. Both bootkube, kvo do this.

euank: I'll take type-safety and autocompletion. For readability lemme adjust the github tabwidth for you real quick...

euank: https://github.com/coreos/container-linux-update-operator/pull/64 should make it more readable.

You can also add `?ts=4` in the url for the same effect.

PR #68, opened 2017-04-26
Bump release script

euank:
Seems like something nice to have.

The remaining bit for doing a release is finishing the jenkins pipeline that pulls a named tag and then does a build-push


build/bump-release
```diff
@@ -0,0 +1,40 @@
...
+
+echo "${VERSION}" > VERSION
+git add VERSION
+git commit -m "version: bump to ${1}"
+git tag --sign "${1}" -m "${1}"
+
+# Put a minor version bump in for the 'dev' version
+echo "${NEXT_VERSION}-dev" > VERSION
+git add VERSION
+git commit -m "version: begin development on v${NEXT_VERSION}"
```
crawford: You use "X.Y.Z" above, but "vX.Y.Z" here.

euank: It is a little confusing.

$1 has v1.2.3
$VERSION has 1.2.3
$NEXT_VERSION has 1.3.0

... lemme fix it so NEXT_VERISON 0's out the patch too actually

euank: switched to not using $1 for anything so it all looks more consistent.

PR #70, opened 2017-04-27
Add more detail on agent

jamiehannaford:
The current README references a non-existent file in `examples`, so I've pointed to the right file. I've also added clarification on how the agent runs and other minor things 😃

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,9 @@
 # Container Linux Update Operator
 
-Container Linux Update Operator is a node reboot controller for Kubernetes on Container Linux Distro.
-When a reboot is needed after updating the system via [update_engine](https://github.com/coreos/update_engine),
-the operator will drain the node before rebooting it.
+Container Linux Update Operator is a node reboot controller for Kubernetes running
+Container Linux images. When a reboot is needed after updating the system via
+[update_engine](https://github.com/coreos/update_engine), the operator will
+drain the node before rebooting it.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -13,29 +14,37 @@ on the node before rebooting.
...
+- The `locksmithd.service` systemd unit on each machine should be masked and stopped in systemd
+
+To unmask a service, run `systemctl unmask <name>`.
+To enable a service, run `systemctl enable <name>`.
+To start/stop a service, run `systemctl start <name>` or `systemctl stop <name>` respectively.
 
 ## Usage
 
-To start `update-operator` and `update-agent`:
+To start the `update-operator` DaemonSet, run:
```
euank: This one's a deployment

README.md
```diff
@@ -13,29 +14,37 @@ on the node before rebooting.
...
 
 ```
-# Open examples/components.yaml and edit the image tag.
-kubectl create -f examples/components.yaml
+kubectl create -f examples/update-operator.yaml
```
euank: Oh, also, should the comment still be there?
Right now it still has "${TAG}", though I'd be in favour of updating TAG to be the latest release instead as an alternate solution.

jamiehannaford: Yeah, I like the idea of using `latest` by default since I assume that's what 99% of users will want.

jamiehannaford:
Thanks for reviewing @euank. I've made the suggested changes.

PR #71, opened 2017-05-02
operator/build: fix tagging consistency

euank:
cc @dghubble, this is the proper fix for the problem we ran into on the installer.
Specifically, the first commit.

The other two are build improvements that are somewhat related, but not all that much.

### DIFF
--- a/pkg/operator/agent_manager.go
+++ b/pkg/operator/agent_manager.go
@@ -220,7 +220,7 @@ func agentDaemonsetSpec(repo string) *v1beta1.DaemonSet {
 }
 
 func agentImageName(repo string) string {
-	return fmt.Sprintf("%s:%s", repo, version.Version)
+	return fmt.Sprintf("%s:v%s", repo, version.Version)
 }
 
 func agentCommand() []string {


scripts/jenkins/Jenkinsfile.release
```diff
@@ -7,6 +7,9 @@ properties([
         string(name: 'TAG',
                defaultValue: 'v0.0.0',
                description: 'CLUO tag to release'),
+        booleanParam(name: "PUSH_LATEST",
```
dghubble: What's the rationale for this? Can we just decide to always have a "latest" tag that is the most recent published version or just never have one.

euank: The reason to not push latest is e.g. if you push a release branch for an older version.

Say we have `1.2.0` and `1.3.0`, and you then release `1.3.1` but need to backport `1.2.1` as well. In that case, `:latest` should not be changed.

euank: This would also be true for e.g. '-alpha` releases

dghubble: Then let's not have latest at all. We don't really want real clusters using the "latest" tag and for development we can build locally.

euank: That was the original plan, see https://github.com/coreos/container-linux-update-operator/pull/70#pullrequestreview-35125422 for why this changed

dghubble: I see. While he's right that many users *want* to deploy latest and I think we *should* require those users to pick an actual hash tag.

Long ago (1.3.X era), a Tectonic release included a component with a "latest" tag by accident. That image then had to be whitelisted and never changed again to avoid breaking clusters at that release. In postmortem, we decided to stop publishing images using "latest" to avoid risk of this happening again. Clusters were to use pinned image versions for best practice, to avoid breaks, so we could freely iterate forward, and we'd receive better bug reports.

I tend to agree with that conclusion and favor not having it.

euank: You're right and I generally agree, but the ship has sailed. at this point ☹️

dghubble: Ok, I still don't think that means we're committed to updating a "latest" forever though - we can stop at any point and earlier is better.

euank: If you want to make a PR moving the example to use the latest published version I can delete that commit from this happily enough.

dghubble: Ok

dghubble: https://github.com/coreos/container-linux-update-operator/pull/73

PR #73, opened 2017-05-08
examples: Switch example deployment to use release tags

dghubble:
Discussion: https://github.com/coreos/container-linux-update-operator/pull/71#pullrequestreview-36907630, https://github.com/coreos/container-linux-update-operator/pull/70#pullrequestreview-35125422


dghubble:
Added to the bump-release script to update the manifest if you can review that again.

PR #85, opened 2017-06-19
make: simplify binary build and test targets

dghubble:
* bin/% should only build the specified target binary, not `./cmd/...`
* Remove `precompile` target, it doesn't seem to be used anywhere
* Remove `integration-test` target, it doesn't do anything
* Remove targets which assume the cwd(`.`)
* Set CGO_ENABLED=0 and REPO at the top to reduce clutter
* Drop `-tags netgo` flag, this shouldn't be needed since Go 1.5


dghubble:
Prefix each line item with `@` to hide the command as well?

PR #86, opened 2017-06-19
glide.yaml: set missing version pins in glide.yaml

dghubble:
Running `make vendor` on a clean checkout produces a diff. With dependencies pinned in glide.yaml, there should only be a diff when a dependency is explicitly updated.

* Add missing version pins to glide.yaml
* glide.yaml lists github.com/golang/mock as a testImport. However, glide is treating it as a regular dependency since it is used in source `pkg/k8sutil/mocks/node_interface_mock.go`. List it as a dependency so that it can be pinned
* Set the k8s.io/kubernetes version explicitly, based on glide.lock

No changes to `vendor` contents, but now running `make vendor` produces no diff.


glide.yaml
```diff
@@ -34,12 +35,13 @@ import:
...
 - package: k8s.io/kubernetes
-  version: ~v1.6.0
+  version: 477efc3cbe6a7effca06bd1452fa356e2201e1ee
   subpackages:
   - pkg/client/leaderelection
   - pkg/client/leaderelection/resourcelock
   - pkg/client/clientset_generated/internalclientset/typed/core/internalversion
   - pkg/kubelet/types
 - package: github.com/golang/glog
-testImport:
```
euank: Was removing `testImport` intentional?

dghubble: Yes, in the description I mentioned glide knows the mocks package is needed in source code. So the package is treated as a normal import. If you try to pin it, it will notice the import and test import are in conflict. Fix is to list the mock package as a regular import and pin it there. `testImport` is then no longer needed

euank:
@dghubble those comments should have been in the commit comment instead of, or in addition to, the github comment for `git log` posterity.

Since the `mocks` package is only important from `_test` files, glide really should be fine with it, but if it's not then oh well.

Issue #75, opened 2017-05-18
Crashing periodically in Tectonic 1.6.2


robszumski:
This doesn't seem normal:

```
$ kubectl get pods
NAME                                              READY     STATUS    RESTARTS   AGE
container-linux-update-agent-ds-fs5c6             1/1       Running   195        5d
container-linux-update-agent-ds-j0bz3             1/1       Running   197        5d
container-linux-update-agent-ds-shcw0             1/1       Running   205        5d
container-linux-update-agent-ds-sj9ds             1/1       Running   190        5d
container-linux-update-agent-ds-vv8w2             1/1       Running   195        5d
```

Logs from a running instance look ok:

```
$ kubectl logs container-linux-update-agent-ds-fs5c6
I0518 18:08:12.158135       1 main.go:42] /bin/update-agent running
I0518 18:08:12.158212       1 agent.go:117] Setting info labels
I0518 18:08:12.195741       1 agent.go:123] Marking node as schedulable
I0518 18:08:12.208552       1 agent.go:134] Setting annotations map[string]string{"container-linux-update.v1.coreos.com/reboot-in-progress":"false", "container-linux-update.v1.coreos.com/reboot-needed":"false"}
I0518 18:08:12.225767       1 agent.go:148] Waiting for ok-to-reboot from controller...
I0518 18:08:12.225865       1 agent.go:205] Beginning to watch update_engine status
I0518 18:08:12.226547       1 agent.go:68] Updating status
```

Logs from a previous run: 

```
$ kubectl logs container-linux-update-agent-ds-fs5c6 -p
I0518 17:28:51.269685       1 main.go:42] /bin/update-agent running
I0518 17:28:51.269748       1 agent.go:117] Setting info labels
I0518 17:28:51.319913       1 agent.go:123] Marking node as schedulable
I0518 17:28:51.336287       1 agent.go:134] Setting annotations map[string]string{"container-linux-update.v1.coreos.com/reboot-in-progress":"false", "container-linux-update.v1.coreos.com/reboot-needed":"false"}
I0518 17:28:51.357811       1 agent.go:148] Waiting for ok-to-reboot from controller...
I0518 17:28:51.357976       1 agent.go:205] Beginning to watch update_engine status
I0518 17:28:51.358551       1 agent.go:68] Updating status
I0518 17:40:40.674321       1 agent.go:68] Updating status
I0518 17:40:50.693834       1 agent.go:68] Updating status
[...]

PR #89, opened 2017-06-21
pkg/agent: avoid exiting on watch termination

euank:
Works around #75. Properly, it would also distinguish between
"retryable" and "unretryable" errors, but client-go doesn't give us
great granularity.

In the future, adding further granularity would be a good idea.

This also tweaks whether delete failures are fatal based on a discussion @aaronlevy and I had a while ago.

cc @dghubble 

Testing done: none yet 😦

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -191,7 +191,8 @@ func (k *Klocksmith) Run() error {
 	for _, pod := range pods {
 		glog.Infof("Terminating pod %q...", pod.Name)
 		if err := k.kc.Pods(pod.Namespace).Delete(pod.Name, deleteOptions); err != nil {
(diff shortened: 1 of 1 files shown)


aaronlevy:
LGTM assuming testing :)

dghubble:
@euank did you push this anywhere already? We can manually test on a cluster, until #90 is addressed.

dghubble:
quay.io/dghubble/container-linux-update-operator:f47bb6f82a7205f601064881492c97dbbb71afd9-dirty (dirty to fix the problems in the build scripts)

dghubble:
I've tested this on a Kubernetes 1.6.4 cluster today (bare-metal, masked locksmithd) for both update-operator and update-agent. Tested fake D-bus signals and real update_engine updates from an older version of stable, no major regressions. I'll try to post something about this testing process soon, to formalize the process.

Issue #90, opened 2017-06-22
Document a strategy for effective development


dghubble:
We should have a process for locally building each component, testing against a remote cluster, and validating a sequence of expected behavior. This should include defining requirements for the test cluster.

- [ ] Local use with a kubeconfig
- [x] Test operator changes independently of agent
- [x] Test agent changes independently of operator
- [ ] Define reqs for agent test given it calls dbus directly
- [x] Define when a successful test run looks like, even if its manual for now

We'd like to get away from a flow where this is tested by editing tectonic-installer.

dghubble:
I can't unassign myself, but this can probably be closed. Its not too practical to develop with a local kubeconfig.

PR #94, opened 2017-06-22
Add development docs and improve Makefile usability

### LINKED ISSUE #92: Improve build-image.sh build strategy
`build-image.sh` builds the `update-operator` and `update-agent` binaries in a build image with rkt. It also builds a container image, assuming sudo-less docker.

* Building the binaries in a uniform environment is a useful target in its own right. I'd like that surfaced in the Makefile
* Building the container image as a target, that doesn't assume sudo-less docker is desirable

dghubble:
* Add docs on development  loop, requirements, and how to validate basic functionality (#90)
* Makefile targets
    * `build/build-image.sh` currently builds a binary, builds an image, and pushes the image behind `make image`. Add Makefile targets for each step individually to make development easier (open to different naming):
   * `make release-bin` - Builds `update-operator` and `update-agent` inside the build image for repeatability
    * `make image` - Builds the docker image
    * `make docker-push` - Pushes the docker image (including to a developer's Quay account)
* Fixes #92 where scripts assumed sudo-less docker was allowed on the developer's machine

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,27 +1,34 @@
-.PHONY:	all bin image clean test vendor
+.PHONY:	all release-bin image clean test vendor
 export CGO_ENABLED:=0
 
-VERSION=$(shell cat VERSION)
+VERSION=$(shell ./build/git-version.sh)
+RELEASE_VERSION=$(shell cat VERSION)
 COMMIT=$(shell git rev-parse HEAD)
 
 REPO=github.com/coreos/container-linux-update-operator
-LD_FLAGS="-w -X $(REPO)/pkg/version.Version=$(VERSION) -X $(REPO)/pkg/version.Commit=$(COMMIT)"
+LD_FLAGS="-w -X $(REPO)/pkg/version.Version=$(RELEASE_VERSION) -X $(REPO)/pkg/version.Commit=$(COMMIT)"
 
-all: bin
+IMAGE_REPO=quay.io/coreos/container-linux-update-operator
 
-bin: bin/update-agent bin/update-operator
+all: bin/update-agent bin/update-operator
 
 bin/%:
 	go build -o $@ -ldflags $(LD_FLAGS) $(REPO)/cmd/$*
 
-image:
-	./build/build-image.sh
-
-clean:
-	rm -rf bin
+release-bin:
+	./build/build-release.sh
 
 test:
 	go test -v $(REPO)/pkg/...
 
+image: release-bin
+	@sudo docker build --rm=true -t $(IMAGE_REPO):$(VERSION) .
+
+docker-push: image
+	@sudo docker push $(IMAGE_REPO):$(VERSION)
+
 vendor:
 	glide update --strip-vendor
+
+clean:
+	rm -rf bin
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ COMMIT=$(shell git rev-parse HEAD)
 REPO=github.com/coreos/container-linux-update-operator
 LD_FLAGS="-w -X $(REPO)/pkg/version.Version=$(RELEASE_VERSION) -X $(REPO)/pkg/version.Commit=$(COMMIT)"
 
-IMAGE_REPO=quay.io/coreos/container-linux-update-operator
+IMAGE_REPO?=quay.io/coreos/container-linux-update-operator
 
 all: bin/update-agent bin/update-operator
 
--- a/build/build-release.sh
+++ b/build/build-release.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -euo pipefail
+set -x
+
+readonly REPO_ROOT=$(git rev-parse --show-toplevel)
+
+sudo rkt run --uuid-file-save=rkt.uuid \
+    --volume repo-root,kind=host,source=${REPO_ROOT} \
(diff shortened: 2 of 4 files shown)


doc/development.md
```diff
@@ -0,0 +1,127 @@
...
+
+In particular, be sure to mask `locksmithd.service` on every Container Linux node.
+
+```
+sudo systemctl mask locksmithd.service
+```
+
+### Deploy
+
+Edit `example/update-operator.yaml` `image` to refer to the development image. Edit the command to tell the operator which development image to use `- --agent-image-repo=quay.io/USERNAME/container-linux-update-operator`.
```
dghubble: Can be removed once #100 merges

build/build-release.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -euo pipefail
+set -x
+
+readonly REPO_ROOT=$(git rev-parse --show-toplevel)
+
+sudo rkt run --uuid-file-save=rkt.uuid \
```
lucab: As much as I like it, should we consider using a multi-stage Dockerfile instead of docker+rkt?

euank: I'm wary of using multi-stage builds still because that requires our Jenkins builder to be running a new enough docker version... which, iirc, some aren't.

dghubble: I left the build process exactly the same, as what was done in `build/build-image.sh`, just without the docker build and docker push commands I had to edit / comment whenever I used it.

dghubble:
Added ^

doc/development.md
```diff
@@ -0,0 +1,127 @@
...
+I0622 18:42:13.840638       1 leaderelection.go:189] successfully acquired lease kube-system/container-linux-update-operator-lock```
+```
+
+Verify a Daemonset `container-linux-update-agent-ds` was created.
+
+```sh
+$ kubectl get daemonsets -n kube-system
+```
+
+Currently, the `update-operator` has two different notions of version, tagged versions like `v0.2.0-dev` and commit versions like the GIT_SHA. `update-operator` assumes if should deploy an `update-agent` of the same tag version, whereas Docker images are tagged with Git tag or commit hash appropriately. Workaround this by adding a Docker tag to your development image `v3.0.0-dev`. The pods in the `container-linux-update-agent` Daemonset will then be able to pull the image.
```
dghubble: This can also be removed once #100 merges. We're moving away from `manage-agent=true`.

doc/development.md
```diff
@@ -0,0 +1,127 @@
...
+
+```
+make image
+```
+
+Push a development image to a personal image repository to test it on a cluster.
+
+```
+export IMAGE_REPO=quay.io/USERNAME/container-linux-update-operator
+make docker-push
```
euank: `make docker-push IMAGE_REPO=quay.io/...` is less of a gotcha than leaving an export lying around.

dghubble: Yeah, I wanted either one to work. We can document just the Makefile flag version if you like

euank: Yeah, I'd prefer that.

dghubble: done

doc/development.md
```diff
@@ -0,0 +1,117 @@
...
+
+In particular, be sure to mask `locksmithd.service` on every Container Linux node.
+
+```
+sudo systemctl mask locksmithd.service --now
+```
+
+### Deploy
+
+Edit `example/update-operator.yaml` and `example/update-agent.yaml` to refer to the development image. Create the `update-operator` deployment and `update-agent` daemonset.
```
sdemos: typo: should be `examples` folder. Also, `examples/update-agent.yaml` doesn't exist in this branch, although I don't think that'll be a problem when it gets merged.

Makefile
```diff
@@ -1,27 +1,34 @@
...
+	./build/build-release.sh
 
 test:
 	go test -v $(REPO)/pkg/...
 
+image: release-bin
+	@sudo docker build --rm=true -t $(IMAGE_REPO):$(VERSION) .
+
+docker-push: image
+	@sudo docker push $(IMAGE_REPO):$(VERSION)
```
sdemos: having this be sudo seems to require me to run `sudo docker login quay.io` before it will successfully push to my repository. Is there a reason this has to be run with sudo, or can we just run it as our user? Same with above.

dghubble: Hm, this is bothersome. Leaving off sudo assumes the user allows `docker` commands to run without sudo (I don't) but I guess the opposite is true for if you do allow it. Was that a one time prompt or is it every time?

doc/development.md
dghubble: Rebased, which should resolve this

Makefile
sdemos: it actually didn't prompt me to login, it just rejected my push saying I wasn't authorized. I didn't figure out what I needed to do until I looked at the makefile and saw sudo.

dghubble: Ok, once you've logged in does it work though? I want us to find a common flow that doesn't require either of us editing the makefile to be productive

sdemos: yeah, once I log in it's fine. I don't mind sudoing everytime, but we should document that it's necessary to `sudo docker login` before you can `make docker-push`.

dghubble: Added
> You may need to run `sudo docker login quay.io` if you haven't already.
> Switch your personal repository to be public so images can be pulled.

PR #103, opened 2017-06-29
vendor: Use glide-vc to strip unused dependencies

dghubble:
Development requirement of `glide-vc` was already documented in #94. I noticed this project wasn't yet using it.


dghubble:
Rebased

Issue #98, opened 2017-06-26
Design: Separate reboot coordination and version operator


dghubble:
### Proposal

Simplify the `update-operator` so that it just performs update coordination (not deploying the agent). Write a separate `version-operator` which ensures the desired version of the coordinator and agent are running (creating them if not, managing migrations if any in future).

In the desired architecture, there would be three single-purpose components (names are examples):

* `container-linux-update-coordinator` - deployment which watches node annotations and coordinates reboots as needed to ensure not too many are rebooting at once. (doesn't deploy agent)
* `container-linux-update-agent` - daemonset which listens for D-Bus signals from `update-engine` and indicates via annotations that it needs a reboot.
* `container-linux-version-operator` - (optional) ensures the correct version of the `update-coordinator` and `update-agent` are running via a reconciliation loop. Handles migrations, if needed.

A user may choose to deploy just the `update-coordinator` and `update-agent`. Or he/she could deploy the `container-linux-version-operator` to manage those apps on their behalf and contain reconciliation and migration logic. 

### Problem

In the current design, the `update-operator` is doing both update coordination and deployment of the agent (conditionally depending on flags) at a hardcoded version. This is problematic

* Users ask for the ability to customize the update-agent daemonset through the update-operator. This expands the scope of the operator greatly and makes it take configuration for the agent it deploys and for coordinating reboots. There is no plan for how migrations will work.
* `update-operator` is not currently performing a reconciliation loop to ensure the `update-agent` is running. Its just performing a one-time check on startup.
* `update-operator` is making assumptions about compatibility between itself and `update-agent` releases. That complexity should live in a component designed for it.
* Oddity or code smell is apparent from the `-manage-agent=true/false` option. https://github.com/coreos/container-linux-update-operator/blob/master/cmd/update-operator/main.go#L19
* Conceivably we only need to update one component, not both. Right now we bump both.
* Differs from practices established through experience in other CoreOS and Tectonic operators.

Discussed with @aaronlevy and @euank

dghubble:
With this proposal, users of this open-source repository would just deploy the two components `container-linux-update-coordinator` deployment and `container-linux-update-agent`. Example manifests would be provided. `--manage-agent` would be removed. Customization would be possible just by editing the manifests yourself.

`container-linux-version-operator` would be written internally with a different audience in mind. It aims to automate creation and migrations of those two components and may have Tectonic-specific logic similar to existing internal operators.

[...]

Issue #76, opened 2017-05-25
CLUO agent should only run on desired nodes


robszumski:
The CLUO agent should not run on anything other than Container Linux nodes. There are a few ways to implement this:

 - CLUO has a node selector on OS type. This would depend on Tectonic adding labels for each OS type.
 - CLUO tolerates a taint that would be added to each CL node. Depends on Tectonic adding that taint.

Note that we set `container-linux-update.v1.coreos.com=coreos` on CL nodes, but I assume this comes _from_ the CLUO, so it can't schedule based on that.

Once we figure out a strategy, let's file an issue on the Tectonic installer.

dghubble:
Adding a node label to nodes and having the daemonset match that would be the way I prefer to go. Tainting nodes would affect other pods we don't intend. This probably needs to be implemented by changing the daemonset the operator creates for the moment - however this is a temporary approach as we intend to decouple the operator and daemonset creation. See #98 or ping directly about those Tectonic discussions.

dghubble:
Discussing this more with @aaronlevy, this is harder than initially stated.

### Goal:

`update-agent` should only run on selected (whitelisted) nodes.  are a number of reasons:

* Heterogeneous clusters have nodes with other OS's
* Container Linux nodes may be diskless, PXE booted images
* Administrator prefers to manage updates through some other mechanism (their prerogative)

It is acceptable for deployment tools (like Tectonic) to label nodes by default (their prerogative).

### Initial Proposal (new users):

Users should simply add a node selector on the `update-agent` daemonset manifest such that `update-agent` only runs on correctly labeled nodes (e.g. `container-linux-update.v1.coreos.com/managed=true`). Set `-manage-agent=false`. *New users in open-source can adopt this strategy now*.

### Migration for Tectonic and people who used manage-agent=true

Tectonic 1.6 used `-manage-agent=true` (now discouraged, see #98 for plans to migrate away from using this) which deployed update-agent on all nodes. Even if we changed the manage-agent behavior to deploy a daemonset with a node selector to match nodes, there is no reliable mechanism to add a new label to nodes on existing clusters (non-starter):

* update-agent adds new whitelist label to nodes it currently runs on, before upgrade. This doesn't work because cloud provider user-data will still launch nodes with the old labels. You _cannot_ upgrade to a CLUO daemonset which switches from running everywhere (blacklist) to selecting nodes (whitelist).
* No components running in Tectonic have the ability to modify user-data

Using Existing Labels:

[...]

PR #104, opened 2017-06-29
cmd/update-operator: Mark manage-agent deprecated

dghubble:
I've updated the example manifests to always set `manage-agent` to false, per design discussions in #98 and #76.

Change `manage-agent` to default to false and show a warning when set to true since this is not the way we want to go going forward. Tectonic has a corresponding change to explicitly request the old behavior until a migration plan is ready. https://github.com/coreos/tectonic-installer/pull/1250

```
W0629 18:35:46.145820       1 main.go:43] use of -manage-agent=true is deprecated and will be removed in the future
I0629 18:35:46.148454       1 main.go:51] /bin/update-operator running
I0629 18:35:46.180891       1 leaderelection.go:179] attempting to acquire leader lease...
```

### DIFF
--- a/cmd/update-operator/main.go
+++ b/cmd/update-operator/main.go
@@ -15,9 +15,10 @@ import (
 
 var (
 	analyticsEnabled = flag.Bool("analytics", true, "Send analytics to Google Analytics")
-	manageAgent      = flag.Bool("manage-agent", true, "Manage the associated update-agent")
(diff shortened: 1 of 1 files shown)


yifan-gu:
@dghubble PR needs a rebase.

dghubble:
Rebased, smaller now.

dghubble:
I guess I will call out, this means Tectonic users running old CLUO who wish to update need to get the update payload which adds the flag to ensure Tectonic continues to use the deprecated mode. Users who just bump the CLUO version by hand without adding the flag will change the behavior. I suppose this is true for any Tectonic upgrade payloads though.

Issue #99, opened 2017-06-27
Add ability to verify health before continuing to next node


thrawn997:
We recently used this operator and it seems like it reboots nodes too quickly - it doesn't wait until pods are ready before proceeding to the next node.  It would be good to either offer a health endpoint we can pass to let the operator know a node is up and ready, or do a check to see if there are any unready pods on the node that was just rebooted.  Right now we intend to write our own controller to check to see if there are any unhealthy pods and then set the paused annotation if there are any unready pods.

lucab:
I think @robszumski mentioned a similar behavior would be useful to handle corner-cases where some heavy skews/incompatibilities happen between CL components (i.e. docker) and k8s versions (too-new OS with too-old k8s, or viceversa). Such a knob would allow to start OS upgrades in such case on some of the node nodes, but stopping it before affecting the cluster as a whole.

thrawn997:
Another example would be the stack crash kernel update. It causes certain java applications to crash. The above proposal of checking for unready pods would limit the impact to a single node as opposed to causing a full outage.

robszumski:
We need to think through if we should simply look at the node itself or pods on it. To be super defensive, I could see a situation where end-user pods cause the update to stop progressing, but that might not be desired. Happy to hear y'alls thoughts.

thrawn997:
Yeah, but it could also enable some use cases that right now stop us from running certain workloads on CoreOS. Imagine running an HA setup of Vault or similar keystore. If a node reboots that is running Vault, it will require users to enter their passwords before it is ready. In the current state, all nodes would reboot without knowing this HA setup is degraded and take down the keystore. Similarly, if you need to have a certain number of database running at a time, but it takes those databases half an hour to warm and become ready, you could easily break the desired state of your cluster.  With this feature all those issues would be solved. You'd just need to alert on the cluster updates being paused until manual intervention. 

One way to resolve this is to know what version the cluster was on, capture what pods are currently not ready, and halt  updates until pods that where ready prior to the update are ready after the update.

dghubble:
During a coordinated auto-update a chosen node should reboot and the operator should wait until that node returns and the update-agent pod starts and changes the node annotations so this encompasses Kubernetes' notion of Node Ready and also shows at least that agent pod was able to run. https://github.com/coreos/container-linux-update-operator/blob/master/pkg/operator/operator.go#L46

[...]

PR #106, opened 2017-06-30
agent: Add pre-reboot hook support

sdemos:
This pr adds support for configuring a systemd unit which should be executed before the agent drains and reboots the node. Any service related to the unit should be configured as `Type=oneshot` or there is no guarantee that the command will be completed before the node is rebooted. The implementation should work whether it is pointed at a target or a oneshot service. 

There is one question that comes up with this implementation. The unit is required to exist, or the agent will exit without rebooting. Should we make this a more graceful failure, or just let the agent fail if the configured target doesn't exist?

### DIFF
--- a/cmd/update-agent/main.go
+++ b/cmd/update-agent/main.go
@@ -14,6 +14,7 @@ import (
 
 var (
 	node         = flag.String("node", "", "Kubernetes node name")
+	unit         = flag.String("unit", "", "systemd unit file to complete before reboot")
 	printVersion = flag.Bool("version", false, "Print version and exit")
 )
 
@@ -34,7 +35,7 @@ func main() {
 		glog.Fatal("-node is required")
 	}
 
-	a, err := agent.New(*node)
+	a, err := agent.New(*node, *unit)
 	if err != nil {
 		glog.Fatalf("Failed to initialize %s: %v", os.Args[0], err)
 	}
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -38,7 +38,7 @@ var (
 	}).AsSelector()
 )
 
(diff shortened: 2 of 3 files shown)


dghubble:
I'd like us to be mindful of how this relates to #99 and the holistic design (which is missing). Piecing it together, we'd like users to be able to define custom "actions" that gate whether a reboot can occur (i.e. block the reboot and rollout of updates) and define custom "actions" which gate whether the update was considered successful (i.e. block update-agent and rollout of updates). By "define actions", that might be specifying a custom systemd unit like in this case or a custom Kubernetes pod which can read and write node labels (probably preferred so w're not encouraging writing app health checking logic into host units). 

Once there is a design to ensure we can facilitate both pre and post reboot customization, at both a systemd and Kubernetes level, it will make it more clear that this PR fulfills a part of that plan.

dghubble:
Rather than configuring each the agent to run a systemd unit via dbus (which solves this specific case only), I'd like to see a design that just adds or expects labels in a defined way.

For users who wish to perform custom actions *before a reboot and drain*, perhaps update-agent can be given a configurable node label to require before actually draining and rebooting. Users can then write any custom logic in a pod (first class) or systemd unit and add the label on success.

For users who wish to perform custom actions *before considering an update successful*, perhaps update-operator can be given an additional configurable node label to require before a node is considered back online (but still mark schedulable). Users can then write any custom logic in a pod or systemd unit and add the label on success. 

I think having a well formed labeling contract provides flexibility over writing this dbus specific code.

sdemos:
I am in the process of redesigning this using kubernetes labels and annotations instead of systemd unit files. I'll open a new pr with those changes once we sort out the details.

PR #107, opened 2017-06-30
pkg/agent: properly stop ReceiveStatuses goroutine

### LINKED ISSUE #93: Shutdown update-agent cleanly, it panics on reboot
When the host is shutting down, `update-agent` will panic. This isn't a huge priority since the host is being shutdown anyway, but users will see it in the logs and worry.

```
0622 20:28:31.776431       1 agent.go:68] Updating status
I0622 20:28:31.776467       1 agent.go:78] Indicating a reboot is needed
I0622 20:28:45.169911       1 agent.go:163] Setting annotations map[string]string{"container-linux-update.v1.coreos.com/reboot-in-progress":"true"}
I0622 20:28:45.190260       1 agent.go:175] Marking node as unschedulable
I0622 20:28:45.208388       1 agent.go:180] Getting pod list for deletion
I0622 20:28:45.228146       1 agent.go:189] Deleting 0 pods
I0622 20:28:45.228252       1 agent.go:199] Node drained, rebooting
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x14252c7]

goroutine 31 [running]:
github.com/coreos/container-linux-update-operator/pkg/updateengine.(*Client).ReceiveStatuses(0xc4202337e0, 0xc420321560, 0xc4203215c0)
        /go/src/github.com/coreos/container-linux-update-operator/pkg/updateengine/client.go:83 +0xf7
created by github.com/coreos/container-linux-update-operator/pkg/agent.(*Klocksmith).watchUpdateStatus
        /go/src/github.com/coreos/container-linux-update-operator/pkg/agent/agent.go:217 +0x138
```

sdemos:
When the node is being rebooted, the ReceiveStatuses goroutine was not being
properly stopped using the stop channel that was provided, which caused a panic
when the agent rebooted the machine.

fixes #93

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -142,7 +142,8 @@ func (k *Klocksmith) Run() error {
 	}
 
 	// watch update engine for status updates
-	go k.watchUpdateStatus(k.updateStatusCallback)
+	watchUpdateStatusStop := make(chan struct{})
+	go k.watchUpdateStatus(k.updateStatusCallback, watchUpdateStatusStop)
 
 	// block until constants.AnnotationOkToReboot is set
 	for {
@@ -154,6 +155,9 @@ func (k *Klocksmith) Run() error {
 		}
 		glog.Warningf("error waiting for an ok-to-reboot: %v", err)
 	}
+
+	// stop watching the update status by closing the channel
+	close(watchUpdateStatusStop)
 
 	// set constants.AnnotationRebootInProgress and drain self
 	anno = map[string]string{
@@ -207,12 +211,11 @@ func (k *Klocksmith) Run() error {
 	return nil
 }
 
-func (k *Klocksmith) watchUpdateStatus(update func(s updateengine.Status)) {
+func (k *Klocksmith) watchUpdateStatus(update func(s updateengine.Status), stop <-chan struct{}) {
 	glog.Info("Beginning to watch update_engine status")
 
 	oldOperation := ""
 	ch := make(chan updateengine.Status, 1)
-	stop := make(chan struct{})
 
 	go k.ue.ReceiveStatuses(ch, stop)
 


pkg/agent/agent.go
```diff
@@ -155,6 +156,9 @@ func (k *Klocksmith) Run() error {
 		glog.Warningf("error waiting for an ok-to-reboot: %v", err)
 	}
 
+	// stop watching the update status
+	watchUpdateStatusStop <- struct{}{}
```
dghubble: You can close the stop channel: `close(watchUpdateStatusStop)`

sdemos: That is in addition to this line, right? I don't think that will send something along the channel to tell it to close the connection.

dghubble: https://blog.golang.org/pipelines#explicit-cancellation is the more usual way of managing done channels. See #108 which attempts to normalize the channel handling in other places as well.

"A receive operation on a closed channel can always proceed immediately, yielding the element type's zero value. This means that [we] can unblock all the receivers simply by closing the done channel. This close is effectively a broadcast signal. We extend each of our functions to accept done as a parameter and arrange for the close to happen via a defer statement, so that all return paths from main will signal the pipeline stages to exit."

dghubble:
Can this be done in conjunction with #110 please? The current channel handling seems atypical and odd.

Operator cmd's should create a top-level done channel (called `stop`) and ensure it is closed via a `defer close(stop)` before the program exists. Components receiving a done channel should pass the done channel on to any goroutines they spawn or create their own if they need to manage them separately. On program exit, closing the top level channel should cause all "descendant" routines to stop.

sdemos:
closing the channel via `defer close(stop)` at the top level wouldn't work in this case since the program never actually exits cleanly, which seems to be what was causing this panic in the first place. It just gets killed while it's waiting for the reboot. I can change this so it closes the channel to indicate completion instead of passing the zero value, but it seems like it would require a more fundamental change to have to top-level `stop` channel work correctly. 

Also, I'm not sure how this and #110 are connected? They make changes to similar functions but the problems they are solving seem unrelated.

dghubble:
Yes, in this case the top-level stop channel is not being used. I'm working on the refactor to pass that separately, no need to wait on it. 

Here the same pattern applies. `Run` launches a goroutine to start the watch and passes in a done channel. It happens to wait for the side effect of running that goroutine and once satisfied, the goroutine can be stopped via `close(stop)`.

This change is a positive improvement because we avoid leaving the goroutine running after we've obtained the results we need. I don't think it solves #93 though - when the program is signaled to shutdown, there is still no mechanism that stops the launched goroutines. That'll be done by the refactor I mentioned.

Sure, let's leave #110 separate.

dghubble:
In short, please close the channel, then I think this is ok.

dghubble:
Can you be sure to leave #93 open?

sdemos:
Sure. I'll leave a comment on it that explains why it is still there too.

PR #108, opened 2017-06-30
pkg/operator: Improve controller structure

dghubble:
Kubernetes [controllers](https://godoc.org/k8s.io/client-go/tools/cache#Controller) and custom operators have a common pattern. It encourages clean shutdowns by passing in a stop channel that is respected by the entire controller. It isolates the reconcilition loop so it can use the `wait` utilities, respects a stop channel, and calls a separate process loop function or method which can be tested by itself.

```
New(config Config) (*Kontroller)

func (k *Kontroller) Run(stop <-chan struct{}) {
  // starting
  wait.Until(processLoop, period, stop)
  // stopping
}

func (k.Kontroller) process() {
  // reconciliation logic
}
```

Usage

```
stop := make(chan struct{})
defer close(stop)
operator.Run(stop)
```

Let's adjust the update operator to get the same benefits.

- Add a `Run(stop <- chan struct{})` to start operator reconciliation process and run until the stop channel is closed
- Move the for loop to a `process()` method which contains the reconciliation logic to coordinate reboots
- Use a simple `wait.Until(process-loop, 30 second period, stop channel) to repeat the reconciliation loop. Open to different times. Flow control dynamic backoff is more complex to reason about and not needed`
- A top-level stop channel is a stop toward solving #93 and other shutdown / cleanup / deadlock issues caused by the lack of a top-down way to stop goroutines 

Tested: `quay.io/dghubble/container-linux-update-operator:ad41ec0dee1b91c8a39c4ef74850caffe90be7ea`

### DIFF
--- a/cmd/update-operator/main.go
+++ b/cmd/update-operator/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"flag"
 	"fmt"
 	"os"
@@ -38,7 +37,10 @@ func main() {
 		analytics.Enable()
 	}
 
-	o, err := operator.New()
+	o, err := operator.New(operator.Config{
+		ManageAgent:    *manageAgent,
+		AgentImageRepo: *agentImageRepo,
+	})
 	if err != nil {
 		glog.Fatalf("Failed to initialize %s: %v", os.Args[0], err)
 	}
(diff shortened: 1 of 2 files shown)


dghubble:
Actually, changing the loop period wasn't my intent, I'm happy to use whatever value. I must have miscalculated how the flow control code was adjusting that period. All the more reason to declare one.

dghubble:
Discussed in real life. Sounds like 30 seconds is quite reasonable and fine.

PR #109, opened 2017-07-05
pkg/updateengine: expose AttemptUpdate method

squeed:


squeed:
@dghubble can I get a review?

pkg/updateengine/status.go
```diff
@@ -18,6 +18,18 @@ import (
 	"fmt"
 )
 
+// The possible update statuses returned from the update engine
+const (
```
dghubble: Include comment with link?

dghubble: I see these are listed (only in comments) in pkg/constants/constants.go as well, though I kinda prefer them here. Pretty sure they correspond to current operation values so maybe specify that as well.

lucab:
@dghubble as the requested change was comment-only and Casey will be away for some time, I just queued a commit on top of his branch adding an external reference. PTAL.

PR #110, opened 2017-07-07
agent: moved getting first ue status to RecieveStatuses func

sdemos:
Generating a status and putting it in our channel would deadlock if there was
already a ue status update waiting for us, which happens when a signal comes
along between the time we subscribe in New() and the time we actually start
consuming status updates. Instead, make the go routine do the first status
signal synthesis for us, and the problem goes away.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -219,14 +219,6 @@ func (k *Klocksmith) watchUpdateStatus(update func(s updateengine.Status), stop 
 
 	go k.ue.ReceiveStatuses(ch, stop)
 
-	// synthesize first message
-	st, err := k.ue.GetStatus()
-	if err != nil {
-		glog.Fatalf("Failed to get update_engine status: %v", err)
-	}
-
-	ch <- st
-
 	for status := range ch {
 		if status.CurrentOperation != oldOperation && update != nil {
 			update(status)
--- a/pkg/updateengine/client.go
+++ b/pkg/updateengine/client.go
@@ -80,7 +80,16 @@ func (c *Client) Close() error {
 	return nil
 }
 
+// ReceiveStatuses receives signal messages from dbus and sends them as Statues
+// on the rcvr channel, until the stop channel is closed. An attempt is made to
+// get the initial status and send it on the rcvr channel before receiving
+// starts.
 func (c *Client) ReceiveStatuses(rcvr chan Status, stop <-chan struct{}) {
+	// if there is an error getting the current status, ignore it and just
+	// move onto the main loop.
+	st, _ := c.GetStatus()
+	rcvr <- st
+
 	for {
 		select {
 		case <-stop:
@@ -105,6 +114,7 @@ func (c *Client) RebootNeededSignal(rcvr chan Status, stop <-chan struct{}) {
 	}
 }
 
+// GetStatus gets the current status from update_engine
 func (c *Client) GetStatus() (result Status, err error) {
 	call := c.object.Call(dbusInterface+".GetStatus", 0)
 	err = call.Err


sdemos:
The tectonic cluster that was consistently producing this problem has been running with this patch over the weekend and is still successfully rebooting machines, so it seems that this fixes the problem.

pkg/agent/agent.go
```diff
@@ -216,14 +216,6 @@ func (k *Klocksmith) watchUpdateStatus(update func(s updateengine.Status)) {
 
 	go k.ue.ReceiveStatuses(ch, stop)
```
dghubble: also `defer close(stop)`

sdemos: This is being taken care of by #107 (at least this specific one). This function currently never stops cleanly so the defered close would never be called anyway.

pkg/agent/agent.go
```diff
@@ -216,14 +216,6 @@ func (k *Klocksmith) watchUpdateStatus(update func(s updateengine.Status)) {
...
 
-	// synthesize first message
```
dghubble: To be more clear, the issue you're solving is that `ch` has been handed off to a goroutine (which may possibly send to that channel). Then we get the status directly and proceed to write it to the channel ourselves which blocks if something was already sent to the channel.

I'm not sure how the original author envisioned this code. Perhaps they intended it without `ch <- st`, the GetStatus call. Still seems odd to me even after your change, since one call is made to GetStatus.

pkg/agent/agent.go
dghubble: Yep, that's even better to use the passed in stop. Let's ad #107 and rebase this on it and test

dghubble: No longer relevant

pkg/agent/agent.go
dghubble: How about triggering the get here, but just not writing to the channel (since that's ReceiveStatuses's job)? It feels odd to do one call upon starting a goroutine with ReceiveStatuses.

```
_, err := k.ue.GetStatus()
```

dghubble:
With these changes, can you publish an image, I'd like to run it real quick on a cluster.

pkg/agent/agent.go
sdemos: `GetStatus()` doesn't actually trigger anything, it's a separate dbus call that gets the current status from updateengine. The information pulled from this call required to make sure that we don't get stuck if updateengine requested a restart when the update agent wasn't listening. It seems to me like there are only two ways we can get the desired behavior in this part; 1. The way it is in this pr or 2. unroll the loop once in `watchUpdateStatus`, but have the unrolled part pull the status information from `GetStatus` instead of from the channel passed to `ReceiveStatuses`.

dghubble: Isn't the idea of `c.ch = make(chan *dbus.Signal, signalBuffer)` and `c.conn.Signal(c.ch)` that `c.ch` receives all signal messages and thus, so does `ReceiveStatus`, which converts that message to a Status and sends it to the `ch` we pass?

If not and we only expect one status, I don't see a reason to loop over a channel, which I think is what you're saying be unrolling it in option 2.

dghubble: As noted during discussion, we can't return as that means we're looping over a channel which can never receive messages, as if the stop channel was closed.

dghubble: Also from discussion, we want to document the the initial special GetStatus call is to solve the specific scenario in which we missed a status message which will not be sent again. The calling code expects to get that message before regular passive receiving of statuses begins.

We plan to ignore any error on the initial GetStatus as there is not much else that can be done. We don't want to retry forever and do want continue on to start receiving statuses. In future I can see us maybe having an `initialFetch bool` option to receive statuses, but a comment would be good for now.

sdemos:
removed error checking and added documentation of decisions to comments.

pkg/updateengine/client.go
```diff
@@ -80,7 +80,16 @@ func (c *Client) Close() error {
 	return nil
 }
 
+// ReceiveStatuses sends status updates from the update_engine along a provided channel.
```
dghubble: nit: Can we re-arrange to the primary focus is still on channel receiving?

"ReceiveStatuses receives signal messages from dbus and sends them as Statues on the rcvr channel, until the stop channel is closed. An attempt is made to get the initial status and send it on the rcvr channel before receiving starts."

PR #111, opened 2017-07-11
Use controller pattern for update-agent

dghubble:
Pass in a top-level stop channel to the agent operator to use when it launches goroutines or calls `time.Sleep`. In theory, if we trapped on signals in cmd, we'd see every goroutine stop and a clean "Stopping agent" log line.

A general controller:

```
New(config Config) (*Kontroller)

func (k *Kontroller) Run(stop <-chan struct{}) {
  // starting
  wait.Until(processLoop, period, stop)
  // stopping
}

func (k.Kontroller) process() {
  // reconciliation logic
}
```

Similar to #108

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -60,59 +60,22 @@ func New(node string) (*Klocksmith, error) {
 	return &Klocksmith{node, kc, nc, ue, lc}, nil
 }
 
-// updateStatusCallback receives Status messages from update engine. If the
-// status is UpdateStatusUpdatedNeedReboot, indicate that with a label on our
-// node.
-func (k *Klocksmith) updateStatusCallback(s updateengine.Status) {
-	glog.Info("Updating status")
-	// update our status
-	anno := map[string]string{
-		constants.AnnotationStatus:          s.CurrentOperation,
-		constants.AnnotationLastCheckedTime: fmt.Sprintf("%d", s.LastCheckedTime),
-		constants.AnnotationNewVersion:      s.NewVersion,
+// Run starts the agent to listen for an update_engine reboot signal and react
+// by draining pods and rebooting. Runs until the stop channel is closed.
+func (k *Klocksmith) Run(stop <-chan struct{}) {
+	glog.V(5).Info("Starting agent")
+
+	// agent process should reboot the node, no need to loop
+	if err := k.process(stop); err != nil {
+		glog.Errorf("Error running agent process: %v", err)
 	}
 
-	// indicate we need a reboot
-	if s.CurrentOperation == updateengine.UpdateStatusUpdatedNeedReboot {
-		glog.Info("Indicating a reboot is needed")
-		anno[constants.AnnotationRebootNeeded] = constants.True
-	}
-
-	wait.PollUntil(10*time.Second, func() (bool, error) {
-		if err := k8sutil.SetNodeAnnotations(k.nc, k.node, anno); err != nil {
-			glog.Errorf("Failed to set annotation %q: %v", constants.AnnotationStatus, err)
-			return false, nil
-		}
-
-		return true, nil
-	}, wait.NeverStop)
+	glog.V(5).Info("Stopping agent")
 }
 
-// setInfoLabels labels our node with helpful info about Container Linux.
-func (k *Klocksmith) setInfoLabels() error {
-	vi, err := k8sutil.GetVersionInfo()
-	if err != nil {
-		return fmt.Errorf("failed to get version info: %v", err)
-	}
-
-	labels := map[string]string{
-		constants.LabelID:      vi.ID,
-		constants.LabelGroup:   vi.Group,
-		constants.LabelVersion: vi.Version,
-	}
-
-	if err := k8sutil.SetNodeLabels(k.nc, k.node, labels); err != nil {
-		return err
-	}
-
-	return nil
-}
-
-// Run runs klocksmithd, reacting to the update_engine reboot signal by
-// draining pods on this kubernetes node and rebooting.
-//
-// TODO(mischief): try to be more resilient against transient failures
-func (k *Klocksmith) Run() error {
+// process performs the agent reconciliation to reboot the node or stops when
+// the stop channel is closed.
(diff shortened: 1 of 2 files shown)


pkg/agent/agent.go
```diff
@@ -142,8 +105,7 @@ func (k *Klocksmith) Run() error {
 	}
 
 	// watch update engine for status updates
-	watchUpdateStatusStop := make(chan struct{})
-	go k.watchUpdateStatus(k.updateStatusCallback, watchUpdateStatusStop)
+	go k.watchUpdateStatus(k.updateStatusCallback, stop)
```
dghubble: Using `stop` (instead of a new channel) means all goroutines spawned fanning out from here can pass the same channel so that all watches can be stopped from the top-level stop channel.

It'd be slighly nicer to have a `merge(stop <-chan struct{}...) stop <-chan struct{}` that was stopped by any of the input channels being stopped since for this particular goroutine we don't care about it after line 160.

pkg/agent/agent.go
```diff
@@ -371,3 +377,15 @@ func (k *Klocksmith) getPodsForDeletion() ([]v1.Pod, error) {
 
 	return pods, nil
 }
+
+// sleepOrDone pauses the current goroutine until the done channel receives
+// or until at least the duration d has elapsed, whichever comes first. This
+// is similar to time.Sleep(d), except it can be interrupted.
+func sleepOrDone(d time.Duration, done <-chan struct{}) {
```
dghubble: Borrowed from the Twitter API https://github.com/dghubble/go-twitter/blob/master/twitter/stream_utils.go#L21

pkg/agent/agent.go
```diff
@@ -206,7 +165,54 @@ func (k *Klocksmith) Run() error {
 	k.lc.Reboot(false)
 
 	// cross fingers
-	time.Sleep(24 * 7 * time.Hour)
+	sleepOrDone(24*7*time.Hour, stop)
+	return nil
```
sdemos: It doesn't seem like this function will ever actually return. `close(stop)` is done by the main function, but only once this function returns, which doesn't happen until the `stop` channel is closed. That doesn't seem like the intended behavior.

dghubble: Hm, what about after 24 hours have passed? sleepOrDone should be just like the time.Sleep call that was there before (followed by reaching the end of the agent behavior, effectively the same as return nil here), except it can be interupted if the top-level stop were closed.

dghubble: The primary reason the agent is unlikely to ever stay in this sleep for the full 24 hours is because a reboot has just been requested. Its kinda unclear what we ever gained by having this sleep, its not like the agent tried again after 24 hours - it just stopped running.

sdemos: Yeah, technically the function will return after 168 hours, but the agent will have almost definitely been killed by this point anyway. This way we get killed if we are doing the right thing and exit cleanly if rebooting has an issue, which seems backward to me. The way this chain of functions is currently called, there is no difference between calling regular sleep and calling sleepOrDone, because stop will only be closed after we wait for 168 hours, stop is closed once `Run` returns and `Run` only returns once `process` returns, and process doesn't return until it sleeps for 168 hours.

dghubble: "stop is closed once Run returns and Run only returns once process returns" while this is technically true, these two things serve different purposes.

The goal of `stop` is if we later add signal catching to main (code change, such as if someone hits ctrl+c). Stopping the channel stops all goroutines or sleeps (if we use sleepOrDone) the agent was doing. The fact that this function returns and Run returns doesn't matter so much as if written for signal catching, the Run is in its own goroutine. Note that its not this way now.

Basically, we should either use an interruptible sleep or (better the more I think about it) decide to just return immedietaly - sleeping seems pointless. Either way, the prior `time.Sleep` should go.

sdemos: We should figure out if there is some reason for this sleep, it seems like something that was put there for a reason. Perhaps the original author was worried it would be restarted by kubernetes before the node was rebooted and screw up the node labels or some other subtle problem. If it isn't needed, however, I totally agree that it shouldn't even exist. (@euank may know more about this?) 

The reason I was pointing out the `stop` channel not being closed is that this will regress on #93. The goroutines started by `process` need to be stopped before the node is told to reboot, or else they will panic.

dghubble: Let's remove both sleep and sleepOrDone. Sound ok?

dghubble: Solving #93 depends on handling the termination signal by closing stop, and thus goroutines and sleeps, like in the above example. It can't depend on Run returning.  From the cleanup angle, neither sleep and sleepOrDone (where currently stop won't be done) are sufficient.

dghubble: Yeah, good point about the Daemonset being rescheduled. We probably would want to leave the sleepOrDone.

dghubble: @sdemos @euank I don't have further changes. I think we want to keep the sleep, but have it be interruptible. #93 is still an issue with or without this change, it requires a future addition to trap the shutdown and close the stop channel to correctly close all routines/sleeps. Resolved to satisfaction?

pkg/agent/agent.go
euank: This has a memory leak, I'll send a PR against the original and expect it to be updated here too I guess.

The memory leak here doesn't really matter since we exit right after and this is called once, but still would be nice not to leave a memory leak lying around

euank: xref https://github.com/dghubble/go-twitter/pull/77

euank: ... and nevermind, I was mixing it up with the ticker issue. The 'after' will be GCd in a week, so no big deal.

pkg/agent/agent.go
sdemos: Yeah, that's fine. We should add the trap soon though.

dghubble:
Lemme add @euank's suggestion. Its non-essential as noted in the Twitter repo, but slightly nicer.

PR #115, opened 2017-07-18
Respect a -kubeconfig and KUBECONFIG path in update-operator

dghubble:
Adding client-go `tools/clientcmd` seems to add 111K lines in vendor. Sorry.

Note: Leader election is still using its own clientset which uses the in-cluster config :-1: As a result, this PR doesn't quite get us able to develop `update-operator` locally with just a `kubeconfig`. I started exploring changing leader election but it got complicated. This PR just switches the main client to get us closer.


dghubble:
Part of why the vendor change is so large is that this project is not pinning dependencies below depth 1 and we probably should.

dghubble:
Step in the right direction or #90

PR #116, opened 2017-07-18
vendor: Remove unused gogo package from glide.yaml

dghubble:
Very minor, but this package is never used.


glide.lock
```diff
@@ -38,7 +38,7 @@ imports:
 - name: github.com/godbus/dbus
   version: 5f6efc7ef2759c81b7ba876593971bfce311eab3
 - name: github.com/gogo/protobuf
-  version: 909568be09de550ed094403c2bf8a261b5bb730a
+  version: e18d7aa8f8c624c915db340349aad4c49b10d173
```
sdemos: if it's not used why is it still in the `glide.lock` file? It looks like the files aren't actually removed from the vendor directory, they were just updated to the latest version now that they aren't being constrained by `glide.yaml`

dghubble: Yeah good point. Something is still using this somewhere in the tree, though glide cleanup says its not. I'll close this, let's not leave it unpinned.

PR #117, opened 2017-07-19
Auto-label Container Linux nodes with update-agent label

dghubble:
*Note: Direct users of CLUO should use `-manage-agent=false` as shown in examples. Deploy the `update-agent` daemonset to select nodes in any way desired. CLUO is not opinionated.*

Background: When the update-operator is run with `manage-agent=true (dprecated)`, the operator creates update-agent as a daemonset across all nodes. We're migrating toward creating a daemonset which has a node selector so admins always have control. #76

When `-manage-agent=true`:

* Auto-label Container Linux nodes which lack the label `container-linux-update.v1.coreos.com/agent` with `container-linux-update.v1.coreos.com/agent=true`
* Don't auto-label nodes running non Container Linux distros
* Admins may set label `container-linux-update.v1.coreos.com/agent=false`
* Ignore nodes once the `container-linux-update.v1.coreos.com/agent` label exists
* A followup migration (external) may switch the update-agent Daemonset to select nodes which have `agent=true` in order to migrate.

### DIFF
--- a/pkg/k8sutil/selector.go
+++ b/pkg/k8sutil/selector.go
@@ -1,7 +1,10 @@
 package k8sutil
 
 import (
+	"strings"
+
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
 	v1api "k8s.io/client-go/pkg/api/v1"
 )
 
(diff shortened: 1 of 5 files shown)


dghubble:
I was using `manage=true` for a while, but it started to feel cumbersome to describe to the reader / user since we already have a `manage-agent` flag that is related but different and "management" has connotations. For now if we use this label for on / off / unset, I felt the label "agent=true" conveyed the agent would run there.

dghubble:
Originally, I'd planned to also add daemonset migration. Following discussion with Tectonic / @abhinavdahiya, Tectonic will has an (internal) plan to take over this responsibility and allow us to delete the agent management code here and retain only this auto-labeling shim. cc @sdemos @aaronlevy @euank

dghubble:
Let's place this behavior behind a completely different flag `-auto-label-container-linux` to support Tectonic. Then `-manage-agent` and `agent-image` can be completely removed.

euank:
@dghubble to be clear, tectonic will run with the new flag set to true, at least for a migration period?

pkg/k8sutil/selector.go
```diff
@@ -18,3 +21,29 @@ func FilterNodesByAnnotation(list []v1api.Node, sel fields.Selector) []v1api.Nod
...
+	return matches
+}
+
+// FilterContainerLinuxNodes filters a list of nodes and returns nodes with a
+// Container Linux OSImage, as reported by the node's /etc/os-release.
+func FilterContainerLinuxNodes(nodes []v1api.Node) []v1api.Node {
+	var matches []v1api.Node
+
+	for _, node := range nodes {
+		if strings.Contains(node.Status.NodeInfo.OSImage, "Container Linux") {
```
euank: nit, but let's do `StartsWith` to be a little more precise.

dghubble:
Yes, for migration, `-auto-label-container-linux` will be set to true and must be kept until there is a suitable replacement. At the same time, the Daemonset will be updated to node select for that this label. `-manage-agent` will then be unused and can be removed.

pkg/k8sutil/selector.go
dghubble: HasPrefix, but added now

PR #120, opened 2017-07-21
examples: Add rolling update strategy and remove label/annotation

dghubble:
* Add update strategy so `kubectl apply` rolling updates if there is an existing daemonset
* By default, update-operator does not manage the update-agent daemonset so the custom label and annotation aren't needed.

### DIFF
--- a/examples/update-agent.yaml
+++ b/examples/update-agent.yaml
@@ -4,13 +4,14 @@ metadata:
   name: container-linux-update-agent
   namespace: kube-system
 spec:
+  updateStrategy:
+    type: RollingUpdate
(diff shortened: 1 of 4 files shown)


examples/update-agent.yaml
```diff
@@ -4,13 +4,14 @@ metadata:
   name: container-linux-update-agent
   namespace: kube-system
 spec:
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 1
```
dghubble: The value shouldn't really matter. 1 is the default.

Issue #50, opened 2017-03-17
Should play nice with "pod disruption budgets"


euank:
This will make deleting things less fragile.

In the case of non-CVE updates, there's not too much risk in trying to fit within such budgets via draining more slowly.

Ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-disruption-budget

wking:
> I'd be happy to steal drain code if it fits.

Instead of copy/pasting this around, it seems like it would be easier to have a centrally-maintained node-draining library.  I've pulled the `kubectl` implementation out into a separate library [here][1], and I'm [trying][2] to get that library landed somewhere under client-go upstream.  If that sounds appealing, help getting the library landed in client-go would be great :).

[1]: https://github.com/wking/kubernetes-drain
[2]: https://github.com/kubernetes/kubernetes/issues/25625#issuecomment-405056988

euank:
Container Linux is deprecated. Closing this issue as stale.

PR #121, opened 2017-07-24
update-operator: Require before and after reboot annotations

sdemos:
This pr adds pre- and post-reboot actions.

With these changes, you will be able to configure actions which get run before a reboot is initiated and after one is completed, effectively gating it on either side. The actions taken are entirely arbitrary, and software that implements them only needs to be able to manipulate kubernetes annotations and read labels.

The changes to the `update-operator` ended up being a little more of an overhaul than I originally intended. With the addition of the new behavior, the `update-operator` has, broadly speaking, four state changes which it is responsible for. Those changes are
1. Find nodes where the post-reboot actions are completed successfully mark everything ready to go
2. Find nodes which have just rebooted and initiate post-reboot actions on them
3. Find nodes where the pre-reboot actions are completed successfully and tell them to reboot
4. Find nodes which want to reboot and initiate pre-reboot actions on them

I broke the process function out into four functions, one for each of those state changes, to make it easier to understand, validate, and modify in the future. 

The `update-agent` has been modified not to mark a node schedulable until the post-reboot actions are complete. Currently, the node continues to be marked as scheduled until it starts draining, which is after the pre-reboot actions are run. It is possible to modify it to mark it as unschedulable while running pre-reboot actions, but it would require more work and I don't think it's necessary to do that. If someone can think of a good reason to mark it unschedulable please disagree with me. 

I will include documentation around the usage in a separate pr, as well as examples.

### DIFF
--- a/cmd/update-operator/main.go
+++ b/cmd/update-operator/main.go
@@ -15,6 +15,8 @@ import (
 )
 
 var (
+	beforeRebootAnnotations flagutil.StringSliceFlag
+	afterRebootAnnotations  flagutil.StringSliceFlag
 	kubeconfig              = flag.String("kubeconfig", "", "Path to a kubeconfig file. Default to the in-cluster config if not provided.")
 	analyticsEnabled        = flag.Bool("analytics", true, "Send analytics to Google Analytics")
 	autoLabelContainerLinux = flag.Bool("auto-label-container-linux", false, "Auto-label Container Linux nodes with agent=true (convenience)")
@@ -25,12 +27,16 @@ var (
 )
 
 func main() {
+	flag.Var(&beforeRebootAnnotations, "before-reboot-annotations", "List of comma-separated Kubernetes node annotations that must be set to 'true' before a reboot is allowed")
+	flag.Var(&afterRebootAnnotations, "after-reboot-annotations", "List of comma-separated Kubernetes node annotations that must be set to 'true' before a node is marked schedulable and the operator lock is released")
+
 	flag.Set("logtostderr", "true")
 	flag.Parse()
 
 	if err := flagutil.SetFlagsFromEnv(flag.CommandLine, "UPDATE_OPERATOR"); err != nil {
 		glog.Fatalf("Failed to parse environment variables: %v", err)
 	}
+
 	// respect KUBECONFIG without the prefix as well
 	if *kubeconfig == "" {
 		*kubeconfig = os.Getenv("KUBECONFIG")
@@ -61,6 +67,8 @@ func main() {
 		AutoLabelContainerLinux: *autoLabelContainerLinux,
 		ManageAgent:             *manageAgent,
 		AgentImageRepo:          *agentImageRepo,
+		BeforeRebootAnnotations: beforeRebootAnnotations,
+		AfterRebootAnnotations:  afterRebootAnnotations,
 	})
 	if err != nil {
 		glog.Fatalf("Failed to initialize %s: %v", os.Args[0], err)
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -81,12 +81,6 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
 		return fmt.Errorf("failed to set node info: %v", err)
 	}
 
-	// we are schedulable now.
-	glog.Info("Marking node as schedulable")
-	if err := k8sutil.Unschedulable(k.nc, k.node, false); err != nil {
-		return err
-	}
-
 	// set coreos.com/update1/reboot-in-progress=false and
 	// coreos.com/update1/reboot-needed=false
 	anno := map[string]string{
@@ -100,6 +94,12 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
 	// Since we set 'reboot-needed=false', 'ok-to-reboot' should clear.
 	// Wait for it to do so, else we might start reboot-looping
 	if err := k.waitForNotOkToReboot(); err != nil {
+		return err
+	}
+
+	// we are schedulable now.
+	glog.Info("Marking node as schedulable")
+	if err := k8sutil.Unschedulable(k.nc, k.node, false); err != nil {
 		return err
 	}
 
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -51,6 +51,11 @@ const (
 	// It is an opaque string, but might be semver.
 	AnnotationNewVersion = Prefix + "new-version"
 
+	// Keys set to true when the operator is waiting for configured annotation
+	// before and after the reboot repectively
+	LabelBeforeReboot = Prefix + "before-reboot"
+	LabelAfterReboot  = Prefix + "after-reboot"
+
 	// Key set by the update-agent to the value of "ID" in /etc/os-release.
 	LabelID = Prefix + "id"
 
--- a/pkg/k8sutil/selector.go
+++ b/pkg/k8sutil/selector.go
@@ -5,8 +5,19 @@ import (
 
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	v1api "k8s.io/client-go/pkg/api/v1"
 )
+
+// NewRequirementOrDie wraps a call to NewRequirement and panics if the Requirment
+// cannot be created. It is intended for use in variable initializations only.
+func NewRequirementOrDie(key string, op selection.Operator, vals []string) *labels.Requirement {
+	req, err := labels.NewRequirement(key, op, vals)
+	if err != nil {
+		panic(err)
+	}
+	return req
+}
 
 // FilterNodesByAnnotation takes a node list and a field selector, and returns
 // a node list that matches the field selector.
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -241,58 +241,104 @@ func (k *Kontroller) withLeaderElection() error {
 func (k *Kontroller) process() {
 	glog.V(4).Info("Going through a loop cycle")
 
-	nodelist, err := k.nc.List(v1meta.ListOptions{})
+	// update nodes which just rebooted
+	err := k.updateJustRebootedNodes()
 	if err != nil {
-		glog.Infof("Failed listing nodes %v", err)
+		glog.Errorf("Failed to update recently rebooted nodes: %v", err)
 		return
 	}
 
-	glog.V(6).Infof("Found nodes: %+v", nodelist.Items)
+	// get all currently rebooting nodes
+	rebootingNodes, err := k.getRebootingNodes()
+	if err != nil {
+		glog.Errorf("Failed to get rebooting nodes: %v", err)
+		return
+	}
+	glog.V(6).Infof("Found %+v rebooting nodes", rebootingNodes)
 
+	// get all rebootable nodes
+	rebootableNodes, err := k.getRebootableNodes()
+	if err != nil {
+		glog.Errorf("Failed to get rebootable nodes: %v", err)
+		return
+	}
+	glog.V(6).Infof("Found %+v rebootable nodes", rebootableNodes)
+
+	// otherwise pick N of the eligible candidates to reboot
+	remainingRebootableCount := maxRebootingNodes - len(rebootingNodes)
+	// markSomeNodesRebootable returns immediately if len(rebootableNodes) == 0
+	k.markSomeNodesRebootable(rebootableNodes, remainingRebootableCount)
+}
+
+func (k *Kontroller) updateJustRebootedNodes() error {
+	nodelist, err := k.nc.List(v1meta.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("Failed listing nodes %v", err)
+	}
+
+	// find nodes which just rebooted
 	justRebootedNodes := k8sutil.FilterNodesByAnnotation(nodelist.Items, justRebootedSelector)
 
 	if len(justRebootedNodes) > 0 {
 		glog.Infof("Found %d rebooted nodes, setting annotation %q to false", len(justRebootedNodes), constants.AnnotationOkToReboot)
 	}
 
+	// for all the nodes which just rebooted, set reboot-ok=true
 	for _, n := range justRebootedNodes {
 		glog.Infof("Setting 'ok-to-reboot=false' for %q", n.Name)
 		if err := k8sutil.SetNodeAnnotations(k.nc, n.Name, map[string]string{
 			constants.AnnotationOkToReboot: constants.False,
 		}); err != nil {
 			glog.Infof("Failed setting annotation %q on node %q to false: %v", constants.AnnotationOkToReboot, n.Name, err)
+			return err
 		}
 	}
 
-	nodelist, err = k.nc.List(v1meta.ListOptions{})
+	return nil
+}
+
+func (k *Kontroller) getRebootableNodes() ([]v1api.Node, error) {
+	nodelist, err := k.nc.List(v1meta.ListOptions{})
 	if err != nil {
-		glog.Infof("Failed listing nodes: %v", err)
-		return
-	}
-
-	// Verify no nodes are still in the process of rebooting to avoid rebooting N > maxRebootingNodes
-	rebootingNodes := k8sutil.FilterNodesByAnnotation(nodelist.Items, stillRebootingSelector)
-	glog.V(6).Infof("Found %+v rebooting nodes", rebootingNodes)
-	if len(rebootingNodes) >= maxRebootingNodes {
-		glog.Infof("Found %d (of max %d) rebooting nodes; waiting for completion", len(rebootingNodes), maxRebootingNodes)
-		for _, n := range rebootingNodes {
-			glog.Infof("Found node %q still rebooting, waiting", n.Name)
-		}
-		return
+		return nil, fmt.Errorf("Failed listing nodes %v", err)
 	}
 
 	rebootableNodes := k8sutil.FilterNodesByAnnotation(nodelist.Items, wantsRebootSelector)
+
+	return rebootableNodes, nil
+}
+
+func (k *Kontroller) getRebootingNodes() ([]v1api.Node, error) {
+	nodelist, err := k.nc.List(v1meta.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("Failed listing nodes %v", err)
+	}
+
+	rebootingNodes := k8sutil.FilterNodesByAnnotation(nodelist.Items, stillRebootingSelector)
+
+	// Verify no nodes are still in the process of rebooting to avoid rebooting N > maxRebootingNodes
+	if len(rebootingNodes) >= maxRebootingNodes {
+		for _, n := range rebootingNodes {
+			glog.Infof("Found node %q still rebooting, waiting", n.Name)
+		}
+		return nil, fmt.Errorf("Found %d (of max %d) rebooting nodes; waiting for completion", len(rebootingNodes), maxRebootingNodes)
+	}
+
+	return rebootingNodes, nil
+}
+
+func (k *Kontroller) markSomeNodesRebootable(rebootableNodes []v1api.Node, remainingRebootableCount int) {
+	// Don't even bother if rebootableNodes is empty. We wouldn't do anything anyway.
 	if len(rebootableNodes) == 0 {
 		return
 	}
 
-	remainingRebootableCount := maxRebootingNodes - len(rebootingNodes)
-	// pick N of the eligible candidates to reboot
 	chosenNodes := make([]*v1api.Node, 0, remainingRebootableCount)
 	for i := 0; i < remainingRebootableCount && i < len(rebootableNodes); i++ {
 		chosenNodes = append(chosenNodes, &rebootableNodes[i])
 	}
 
+	// mark the chosen nodes as rebootable
 	glog.Infof("Found %d nodes that need a reboot", len(chosenNodes))
 	for _, node := range chosenNodes {
 		k.markNodeRebootable(node)
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -8,6 +8,7 @@ import (
 	"github.com/golang/glog"
 	v1meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -63,21 +64,37 @@ var (
 	// it would like to reboot, and false when it starts up.
 	//
 	// If constants.AnnotationRebootPaused is set to "true", the update-agent will not consider it for rebooting.
-	wantsRebootSelector = fields.ParseSelectorOrDie(constants.AnnotationRebootNeeded + "==" + constants.True + "," + constants.AnnotationRebootPaused + "!=" + constants.True)
+	wantsRebootSelector = fields.ParseSelectorOrDie(constants.AnnotationRebootNeeded + "==" + constants.True +
+		"," + constants.AnnotationRebootPaused + "!=" + constants.True +
+		"," + constants.AnnotationOkToReboot + "!=" + constants.True +
+		"," + constants.AnnotationRebootInProgress + "!=" + constants.True)
 
 	// stillRebootingSelector is a selector for the annotation set expected to be
 	// on a node when it's in the process of rebooting
 	stillRebootingSelector = fields.Set(map[string]string{
-		constants.AnnotationOkToReboot:       constants.True,
-		constants.AnnotationRebootNeeded:     constants.True,
-		constants.AnnotationRebootInProgress: constants.True,
+		constants.AnnotationOkToReboot:   constants.True,
+		constants.AnnotationRebootNeeded: constants.True,
 	}).AsSelector()
+
+	// beforeRebootReq requires a node to be waiting for before reboot checks to complete
+	beforeRebootReq = k8sutil.NewRequirementOrDie(constants.LabelBeforeReboot, selection.In, []string{constants.True})
+
+	// afterRebootReq requires a node to be waiting for after reboot checks to complete
+	afterRebootReq = k8sutil.NewRequirementOrDie(constants.LabelAfterReboot, selection.In, []string{constants.True})
+
+	// notBeforeRebootReq and notAfterRebootReq are the inverse of the above checks
+	notBeforeRebootReq = k8sutil.NewRequirementOrDie(constants.LabelBeforeReboot, selection.NotIn, []string{constants.True})
+	notAfterRebootReq  = k8sutil.NewRequirementOrDie(constants.LabelAfterReboot, selection.NotIn, []string{constants.True})
 )
 
 type Kontroller struct {
 	kc kubernetes.Interface
 	nc v1core.NodeInterface
 	er record.EventRecorder
+
+	// annotations to look for before and after reboots
+	beforeRebootAnnotations []string
+	afterRebootAnnotations  []string
 
 	leaderElectionClient        clientset.Interface
 	leaderElectionEventRecorder record.EventRecorder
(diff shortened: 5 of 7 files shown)


pkg/operator/operator.go
```diff
@@ -229,76 +257,204 @@ func (k *Kontroller) withLeaderElection() error {
...
+				glog.Infof("Failed deleting label %q on node %q: %v", constants.LabelRebootAction, n.Name, err)
+				continue
+			}
+
+			// manually call UpdateNodeRetry so that we can make sure the node still wants to reboot
+			glog.Infof("Marking %q ok to reboot", n.Name)
+			err = k8sutil.UpdateNodeRetry(k.nc, n.Name, func(node *v1api.Node) {
+				if !wantsRebootSelector.Matches(fields.Set(node.Annotations)) {
+					glog.Warningf("Node %v no longer wanted to reboot while we were trying to mark it so: %v", node.Name, node.Annotations)
+					return
```
squeed: Should this `return` be a `continue`?

lucab:
I've left some initial comments inline but I'll do a deeper pass at a later point. Regarding your two WIP points: "yes please" to both, and we use cobra externsively in all other projects.

squeed:
Here is a specific use-case for this, for your own inspiration:

When CL is upgraded on a node, the torcx portion of the agent will need to make sure the appropriate docker addon is fetched. This has the following requirements:

* Executed after or during the CL update, before the reboot
* A fetch failure should prevent the reboot and (ideally) prevent the new version from being selected, should a reboot accidentally happen
* After reboot, a GC process should be run.

It would be helpful if the new CL version was an argument, but it's not absolutely necessary

pkg/operator/operator.go
sdemos: This line is part of an anonymous function that starts on line 317. It's a little weird, but this node annotation update needs to make sure the node still wants to reboot before it gets marked to do so.

pkg/action/action.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	v1api "k8s.io/client-go/pkg/api/v1"
+
+	"github.com/coreos/container-linux-update-operator/pkg/constants"
+)
+
+const (
+	// ActionsSucceeded is returned when all actions have set their configured
+	// annotation to true
+	ActionsSucceeded ActionsResult = iota
```
sdemos: I'm going to add an ActionsUnknown value. That way the zero value is not something which is used to represent an actual state, and instead represents the absence of that state. Right now it does the same thing, since everywhere that uses this value just checks `result == ActionsSucceeded`

pkg/agent/agent.go
```diff
@@ -103,6 +97,12 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
 		return err
 	}
 
+	// we are schedulable now.
+	glog.Info("Marking node as schedulable")
+	if err := k8sutil.Unschedulable(k.nc, k.node, false); err != nil {
```
sdemos: That is an interesting question. It seems like we just indiscriminately mark nodes schedulable and not schedulable for our own uses right now, and that is something that is worth thinking about. I do think it's out of scope for this particular change though.

squeed: Derp, sorry. I actually did forget my glasses at home today.

lucab: @sdemos ack, let's move this to its own ticket/PR. Perhaps there is aready one, I didn't check (my bad). Can you take care of that?

sdemos: Other than #122 I haven't seen any other reference to this specific issue, although we have some other issues with being good neighbors (#50, #61).

sdemos:
Found out that flagutil in coreos/pkg has a slice argument flags extension already in it. I'm just using that instead of moving to a new library for now.

cmd/update-operator/main.go
```diff
@@ -15,6 +15,8 @@ import (
 )
 
 var (
+	preRebootActions        flagutil.StringSliceFlag
```
dghubble: these are annotations, not "actions" (let's avoid this term in the real implementation, it doesn't mean much)

cmd/update-operator/main.go
```diff
@@ -25,12 +27,16 @@ var (
 )
 
 func main() {
+	flag.Var(&preRebootActions, "require-pre-reboot", "list of comma-separated kubernetes annotations required to be set to 'true' before the operator will authorize a reboot")
```
dghubble: Drop mention of authorization.

before-reboot-annotations - "List of comma separated Kubernetes node annotations that must be set to true before a reboot is allowed."

cmd/update-operator/main.go
```diff
@@ -25,12 +27,16 @@ var (
...
+	flag.Var(&postRebootActions, "require-post-reboot", "list of comma-separated kubernetes annotations required to be set to 'true' before the operator will consider a reboot successful")
```
dghubble: after-reboot-annotations - "List of comma separated Kubernetes node annotations that must be set to true before a node is marked scheduleable and the operator lock is released."

examples/post-reboot-action.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: extensions/v1beta1
```
dghubble: Examples in a separate PR which has a tutorial and docs.

pkg/action/action.go
```diff
@@ -0,0 +1,62 @@
...
+// ActionsResult is the result of a given set of actions
+type ActionsResult int
+
+// Action is a pre- or post-reboot action to perform.
+type Action struct {
+	annotation string
+}
+
+// New creates a new action with the given annotation
+func New(annotation string) *Action {
```
dghubble: I thought we were using Action as a placeholder phrase in design docs, not as the actual implementation name. Its rather vague. Especially because CLUO is not actuating the actions, its simply waiting for annotation requirements to be fulfilled.

pkg/constants/constants.go
```diff
@@ -51,6 +51,14 @@ const (
 	// It is an opaque string, but might be semver.
 	AnnotationNewVersion = Prefix + "new-version"
 
+	// Key set by the update-operator when pre- and post-reboot actions should run
+	LabelRebootAction = Prefix + "reboot-action"
```
dghubble: Again, I didn't think we were actually exposing "action" in the label names. Why not use something the user would expect like:

```
nodeSelector:
    container-linux-update.v1.coreos.com/before-reboot: "true"
```

or

```
nodeSelector:
    container-linux-update.v1.coreos.com/after-reboot: "true"
```

pkg/operator/operator.go
```diff
@@ -241,76 +262,229 @@ func (k *Kontroller) withLeaderElection() error {
...
-			constants.AnnotationOkToReboot: constants.False,
-		}); err != nil {
-			glog.Infof("Failed setting annotation %q on node %q to false: %v", constants.AnnotationOkToReboot, n.Name, err)
+func (k *Kontroller) checkPreRebootActions() error {
+	nodelist, err := k.nc.List(v1meta.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("Failed listing nodes %v", err)
+	}
+
+	preRebootReq, err := labels.NewRequirement(constants.LabelRebootAction, selection.In, []string{constants.RebootActionPre})
```
dghubble: I added a little helper to construct Requirements at compilation time

pkg/operator/operator.go
```diff
@@ -241,76 +262,229 @@ func (k *Kontroller) withLeaderElection() error {
...
+		glog.Errorf("Failed to check pre-reboot actions: %v", err)
+		return
+	}
 
-	if len(justRebootedNodes) > 0 {
-		glog.Infof("Found %d rebooted nodes, setting annotation %q to false", len(justRebootedNodes), constants.AnnotationOkToReboot)
+	// take some number of the rebootable nodes.
+	// remove any old action annotations and add the reboot-action=pre label.
+	// this starts the pre-reboot actions.
+	err = k.markPreRebootActionable()
```
dghubble: I can see how method names like these developed, but looking at it with fresh eyes, it seems nonsensical. If we went with the nomenclature I mentioned, this becomes `markBeforeReboot()` which adds a before-reboot=true annotation.

pkg/operator/operator.go
```diff
@@ -241,76 +262,229 @@ func (k *Kontroller) withLeaderElection() error {
...
+func (k *Kontroller) checkPostRebootActions() error {
+	nodelist, err := k.nc.List(v1meta.ListOptions{})
 	if err != nil {
-		glog.Infof("Failed listing nodes: %v", err)
-		return
+		return fmt.Errorf("Failed listing nodes %v", err)
 	}
 
-	// Verify no nodes are still in the process of rebooting to avoid rebooting N > maxRebootingNodes
+	postRebootReq, err := labels.NewRequirement(constants.LabelRebootAction, selection.In, []string{constants.RebootActionPost})
```
dghubble: If all parts of the requirement are constants or known upfront, checkout `MustRequirement` and a var declaration.

pkg/operator/operator.go
```diff
@@ -241,76 +262,229 @@ func (k *Kontroller) withLeaderElection() error {
...
+func (k *Kontroller) markPostRebootActionable() error {
+	nodelist, err := k.nc.List(v1meta.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("Failed listing nodes %v", err)
+	}
+
+	// find nodes which just rebooted
+	justRebootedNodes := k8sutil.FilterNodesByAnnotation(nodelist.Items, justRebootedSelector)
+	// also filter out any nodes that are already labeled with reboot-action=post
+	notPostRebootReq, err := labels.NewRequirement(constants.LabelRebootAction, selection.NotIn, []string{constants.RebootActionPost})
```
dghubble: same

pkg/action/action.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// ActionsCompleted inspects a nodes annotations to determine if a set of
+// actions has succeeded, failed, or is pending completion
+// the annotations have three possible states
+//   true - the action has completed successfully
+//   anything else - the action has not yet completed
+// if all of the annotations exist and are true, return ActionsSucceeded
+// if any action does not exist or is not set to true, return ActionsPending
+func ActionsCompleted(node v1api.Node, actions []*Action) ActionsResult {
```
dghubble: In offline conversation we talked about removing this package, removing actions, and just having something kinda like `HasAnnotations(node, annotations []Annotations) bool`

sdemos:
Removed WIP. I'm going to include docs changes in another pr once we are comfortable with these changes, so that I can write them based on the final form. Please review @dghubble @lucab @euank @squeed

pkg/k8sutil/selector.go
```diff
@@ -5,9 +5,20 @@ import (
 
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	v1api "k8s.io/client-go/pkg/api/v1"
 )
 
+// NewRequirementOrDie wraps a call to NewRequirement and panics if the Requirment
+// cannot be created. It is intended for use in variable initializations only.
+func NewRequirementOrDie(key string, op selection.Operator, vals []string) *labels.Requirement {
```
dghubble: In the standard library, they name these functions MustThing like MustTemplate, MustParse, etc. They also wrap the return (Object, error), not the constructor as constructor funcs can change over time.

I prefer you keep the previous MustRequirement function which had this in mind.

pkg/operator/operator.go
```diff
@@ -241,76 +273,213 @@ func (k *Kontroller) withLeaderElection() error {
...
+		label: constants.True,
+	})
+	if err != nil {
+		return fmt.Errorf("Failed setting label %q to %q for node %q: %v", label, constants.True, node, err)
+	}
+
+	return nil
+}
+
+func hasAnnotations(node v1api.Node, annotations []string) bool {
```
dghubble: add method comment about this being an AND or maybe call it hasAllAnnotations

pkg/operator/operator.go
```diff
@@ -241,76 +273,213 @@ func (k *Kontroller) withLeaderElection() error {
...
+		glog.Errorf("Failed to update recently rebooted nodes: %v", err)
+		return
+	}
 
-	justRebootedNodes := k8sutil.FilterNodesByAnnotation(nodelist.Items, justRebootedSelector)
+	// find nodes with the before-reboot=true label and check if all provided
+	// annotations are set. if all annotations are set to true then remove the
+	// before-reboot=true label and set reboot=ok=true, telling the agent it's
+	// time to reboot.
+	err = k.checkBeforeReboot()
```
dghubble: Move the comment down to be the formal method comment. In the process, let's more clearly define the error - the current handling seems unexpected

dghubble: Discussed. Its clear that these 4 state transition functions can run in different orders and the current ordering is coincidental and errors are divided btw serious errors worth waiting for the process loop and those that are more benign. 

I'd favor these methods returning any error that occurs though. Its easier to explain and to write tests if the errors are returned rather than being log lines. Waiting for the process loop shouldn't be too common since the places we can error now already have reties baked in and if we switch to Watches in the future, the fixed period goes away anyway. If we instead plan for them to be 4 independent loops, it'd be ok to spawn 4 goroutines for them, that way nobody starts depending on their ordering.

pkg/k8sutil/selector.go
sdemos: There is a method in the field kubernetes package called `ParseSelectorOrDie` that I was emulating with this call, since we use that one to construct selectors as well. 

As for the explicit arguments vs the return, I felt like it was cleaner to just have one function call vs two, and the type system will trivially catch any issues with modifications to the underlying constructor that we would have to care about in this function.

pkg/operator/operator.go
sdemos: I like the idea of putting them in goroutines to enforce the fact that the ordering doesn't matter, but I feel like that is asking for weird race condition bugs. The ordering of the state checking and modification doesn't matter, but it would take a lot more work to make us confident that the functions are safe to run concurrently. 

I agree with the error returning though. I'll change that right now.

pkg/k8sutil/selector.go
dghubble: sure

sdemos:
A few changes pushed just now - all functions called by process return an error if there is an error with the subcalls, there are doc comments for each of the four main state change functions, and I consolidated the kubernetes node label and annotation updates into a single call for each state changes instead of two update calls in a row, to prevent possible strange half-states if something fails in the middle of being updated (it's also a little faster and imo cleaner looking).

pkg/operator/operator.go
```diff
@@ -241,76 +273,237 @@ func (k *Kontroller) withLeaderElection() error {
...
+
+	if len(justRebootedNodes) > 0 {
+		glog.Infof("Found %d rebooted nodes, setting label %q to %q", len(justRebootedNodes), constants.LabelAfterReboot, constants.True)
 	}
+
+	// for all the nodes which just rebooted, remove any old annotations and add the after-reboot=true label
+	for _, n := range justRebootedNodes {
+		err = k.mark(n.Name, constants.LabelAfterReboot, k.afterRebootAnnotations)
+		if err != nil {
+			return fmt.Errorf("Failed to mark node for after reboot checks: %v", err)
```
dghubble: same

pkg/operator/operator.go
```diff
@@ -241,76 +273,237 @@ func (k *Kontroller) withLeaderElection() error {
...
 }
 
-// markNodeRebootable adds ok to reboot annotations to the given node.
-func (k *Kontroller) markNodeRebootable(n *v1api.Node) {
-	glog.Infof("Marking %q ok to reboot", n.Name)
-	if err := k8sutil.UpdateNodeRetry(k.nc, n.Name, func(node *v1api.Node) {
-		if !wantsRebootSelector.Matches(fields.Set(node.Annotations)) {
-			glog.Warningf("Node %v no longer wanted to reboot while we were trying to mark it so: %v", node.Name, node.Annotations)
-			return
+func (k *Kontroller) mark(n string, label string, annotations []string) error {
```
dghubble: Can this take the `node v1api.Node` instead of `n string` please or change n to nodeName

pkg/operator/operator.go
```diff
@@ -241,76 +273,237 @@ func (k *Kontroller) withLeaderElection() error {
...
+	err := k.checkAfterReboot()
 	if err != nil {
-		glog.Infof("Failed listing nodes %v", err)
+		glog.Errorf("Failed to check after reboot: %v", err)
 		return
 	}
 
-	glog.V(6).Infof("Found nodes: %+v", nodelist.Items)
+	// find nodes which just rebooted but haven't run after-reboot checks.
+	// remove any old annotations and add the after-reboot=true label.
```
dghubble: "any old annotations" -> before-reboot annotations

Replace throughout

pkg/operator/operator.go
```diff
@@ -241,76 +273,237 @@ func (k *Kontroller) withLeaderElection() error {
...
+	glog.V(4).Info("Marking rebooted nodes with after-reboot label")
```
dghubble: nit: Can we consistently say "Labeling ..." or "Annotating ..." in log messages?

pkg/operator/operator.go
```diff
@@ -241,76 +273,237 @@ func (k *Kontroller) withLeaderElection() error {
...
+
+	for _, n := range preRebootNodes {
+		if hasAllAnnotations(n, k.beforeRebootAnnotations) {
+			glog.V(4).Infof("Deleting label %q for %q", constants.LabelBeforeReboot, n.Name)
+			glog.V(4).Infof("Setting annotation %q to true for %q", constants.AnnotationOkToReboot, n.Name)
+			err = k8sutil.UpdateNodeRetry(k.nc, n.Name, func(node *v1api.Node) {
+				delete(node.Labels, constants.LabelBeforeReboot)
+
+				// make sure that the node still wants to reboot
+				if !wantsRebootSelector.Matches(fields.Set(node.Annotations)) {
```
dghubble: We'd also want to cleanup before-reboot annotations that may have been applied. Putting before-reboot and after-boot labels on a node kinda starts the window of time where custom pods add annotations - seems like we should promise to run a cleanup method at the end of that window (i.e. when removing the before-reboot and after-reboot labels).

pkg/operator/operator.go
```diff
@@ -241,76 +273,237 @@ func (k *Kontroller) withLeaderElection() error {
...
+	// find nodes which just rebooted
+	justRebootedNodes := k8sutil.FilterNodesByAnnotation(nodelist.Items, justRebootedSelector)
+	// also filter out any nodes that are already labeled with after-reboot=true
+	justRebootedNodes = k8sutil.FilterNodesByRequirement(justRebootedNodes, notAfterRebootReq)
+
+	if len(justRebootedNodes) > 0 {
+		glog.Infof("Found %d rebooted nodes, setting label %q to %q", len(justRebootedNodes), constants.LabelAfterReboot, constants.True)
 	}
+
+	// for all the nodes which just rebooted, remove any old annotations and add the after-reboot=true label
```
dghubble: by any old annotations, you mean after reboot annotations from a prior post reboot window?

pkg/operator/operator.go
```diff
@@ -241,76 +273,237 @@ func (k *Kontroller) withLeaderElection() error {
...
+	}
+
+	postRebootNodes := k8sutil.FilterNodesByRequirement(nodelist.Items, afterRebootReq)
+
+	for _, n := range postRebootNodes {
+		if hasAllAnnotations(n, k.afterRebootAnnotations) {
+			glog.V(4).Infof("Deleting label %q for %q", constants.LabelAfterReboot, n.Name)
+			glog.V(4).Infof("Setting annotation %q to false for %q", constants.AnnotationOkToReboot, n.Name)
+			err = k8sutil.UpdateNodeRetry(k.nc, n.Name, func(node *v1api.Node) {
+				delete(node.Labels, constants.LabelAfterReboot)
```
dghubble: Similar to above, where do we wanna cleanup the after-reboot annotations? I'm missing that part

pkg/operator/operator.go
sdemos: The advantage of checking if it no longer wants to reboot here is that it checks at the last possible second, so there isn't a possibility of race conditions. We can also check it in the main loop, but we should still continue to check it here I think.

sdemos: As far as cleaning up before-reboot and after-reboot annotations, it's currently doing it right before it applies the respective label to the node, to ensure an entirely clean environment, and that there aren't any old annotations lying around from the last run. It seems to me like if we only clean them up when we remove the label after completion, there is a possibility that a stray annotation might make it into the mix. Maybe I'm being paranoid.

dghubble: My main worry is that if this inconsistency can occur here, it can occur at other times too so we should patch it up in general reconciliation

dghubble: > isn't a possibility of race conditions

I mean, its still best effort here. Checking at the last second still means the live state can change before you put the label on. We're reducing the likelihood

pkg/operator/operator.go
sdemos: yes

dghubble: what about the idea we call one cleanupAnnotations method any place in code we remove the before-reboot or after-reboot label? That way the user's nodes don't have annotations lying around once the node is happily in steady state

dghubble: and sure, you could still call cleanup here to be defensive and not assume they were cleanup on the prior run

pkg/operator/operator.go
```diff
@@ -241,76 +273,237 @@ func (k *Kontroller) withLeaderElection() error {
...
 	}
 
+	// set before=reboot=true for the chosen nodes
 	glog.Infof("Found %d nodes that need a reboot", len(chosenNodes))
-	for _, node := range chosenNodes {
-		k.markNodeRebootable(node)
+	for _, n := range chosenNodes {
+		err = k.mark(n.Name, constants.LabelBeforeReboot, k.beforeRebootAnnotations)
+		if err != nil {
+			return fmt.Errorf("Failed to mark node for before reboot checks: %v", err)
```
sdemos: this is just to give context of if it's doing the before-reboot or after-reboot checks. That information is provided in the error returned by `mark`. I'll make sure the error in `mark` includes the label name though.

pkg/operator/operator.go
sdemos: yeah, I suppose that's true. We can add a function that makes sure everything is still in a valid state, ie we shouldn't have the `before-reboot` label if `reboot-needed=false`. That would bring `checkBeforeReboot` more in line with `checkAfterReboot`, which would be nice.

pkg/operator/operator.go
```diff
@@ -241,76 +273,237 @@ func (k *Kontroller) withLeaderElection() error {
...
+
+// markBeforeReboot gets nodes which want to reboot and marks them with the
+// before-reboot=true label. This is considered the beginning of the reboot
+// process from the perspective of the update-operator. It will only mark
+// nodes with this label up to the maximum number of concurrently rebootable
+// nodes as configured with the maxRebootingNodes constant.
+// If there is an error getting the list of nodes or updating any of them, an
+// error is immediately returned.
+func (k *Kontroller) markBeforeReboot() error {
+	nodelist, err := k.nc.List(v1meta.ListOptions{})
```
sdemos: I have some ideas around that, but I think that for now we should just keep the list call in this function and maybe change it later.

pkg/operator/operator.go
```diff
@@ -241,76 +273,284 @@ func (k *Kontroller) withLeaderElection() error {
...
 	glog.Infof("Found %d nodes that need a reboot", len(chosenNodes))
-	for _, node := range chosenNodes {
-		k.markNodeRebootable(node)
+	for _, n := range chosenNodes {
+		err = k.mark(n.Name, constants.LabelBeforeReboot, k.beforeRebootAnnotations)
+		if err != nil {
+			return fmt.Errorf("Failed to mark node for before reboot checks: %v", err)
+		}
+		glog.Infof("Waiting for configured before-reboot annotations on node %q", n.Name)
+		glog.V(4).Infof("Waiting for the following annotations: %v", k.beforeRebootAnnotations)
```
dghubble: This log line doesn't make sense when people aren't using this feature:

```
Waiting for configured before-reboot annotations on node 10.132.109.209
```

pkg/operator/operator.go
```diff
@@ -241,76 +273,284 @@ func (k *Kontroller) withLeaderElection() error {
...
+	}
+
+	// find nodes which are still rebooting
 	rebootingNodes := k8sutil.FilterNodesByAnnotation(nodelist.Items, stillRebootingSelector)
-	glog.V(6).Infof("Found %+v rebooting nodes", rebootingNodes)
+	// nodes running before and after reboot checks are still considered to be "rebooting" to us
+	beforeRebootNodes := k8sutil.FilterNodesByRequirement(nodelist.Items, beforeRebootReq)
+	rebootingNodes = append(rebootingNodes, beforeRebootNodes...)
+	afterRebootNodes := k8sutil.FilterNodesByRequirement(nodelist.Items, afterRebootReq)
+	rebootingNodes = append(rebootingNodes, afterRebootNodes...)
```
dghubble: Changes to the logic here seem to cause a node to get stuck being considered to be rebooting. Investigating.

dghubble:
```
I0803 22:04:43.194547       1 operator.go:515] Found 1 rebooted nodes, setting label "container-linux-update.v1.coreos.com/after-reboot" to "true"
I0803 22:04:43.301646       1 operator.go:524] Waiting for configured after-reboot annotations on node "10.132.109.209"
I0803 22:04:43.696275       1 operator.go:457] Found node "10.132.109.209" still rebooting, waiting
I0803 22:04:43.696338       1 operator.go:459] Found 1 (of max 1) rebooting nodes; waiting for completion
```
Actually, I think this is just the oddity of the logging.

dghubble: I don't think this is a new issue, but we should indicate the lack of rebooting nodes if we're going to info log the count of rebooting nodes. It gives the appearance of being stuck at 1.

pkg/operator/operator.go
dghubble: Can we just be consistent in the logs and say "annotating node" or "labeling node". I'm mainly objecting to "mark" since it doesn't mean anything to a Kubernetes user

pkg/operator/operator.go
```diff
@@ -241,76 +273,284 @@ func (k *Kontroller) withLeaderElection() error {
...
+	// set before-reboot=true for the chosen nodes
 	glog.Infof("Found %d nodes that need a reboot", len(chosenNodes))
-	for _, node := range chosenNodes {
-		k.markNodeRebootable(node)
+	for _, n := range chosenNodes {
+		err = k.mark(n.Name, constants.LabelBeforeReboot, k.beforeRebootAnnotations)
+		if err != nil {
+			return fmt.Errorf("Failed to mark node for before reboot checks: %v", err)
+		}
+		glog.Infof("Waiting for configured before-reboot annotations on node %q", n.Name)
```
dghubble: Let's just merge these two log lines and list the annotations in the normal case. Also, you can drop "configured", its assumed

pkg/operator/operator.go
```diff
@@ -241,76 +273,284 @@ func (k *Kontroller) withLeaderElection() error {
...
+
+	for _, n := range postRebootNodes {
+		if hasAllAnnotations(n, k.afterRebootAnnotations) {
+			glog.V(4).Infof("Deleting label %q for %q", constants.LabelAfterReboot, n.Name)
+			glog.V(4).Infof("Setting annotation %q to false for %q", constants.AnnotationOkToReboot, n.Name)
+			err = k8sutil.UpdateNodeRetry(k.nc, n.Name, func(node *v1api.Node) {
+				delete(node.Labels, constants.LabelAfterReboot)
+				// cleanup the after-reboot annotations
+				for _, annotation := range k.afterRebootAnnotations {
+					delete(node.Annotations, annotation)
```
dghubble: Verbose log these deletions

pkg/operator/operator.go
```diff
@@ -241,76 +273,284 @@ func (k *Kontroller) withLeaderElection() error {
...
+
+	for _, n := range preRebootNodes {
+		if hasAllAnnotations(n, k.beforeRebootAnnotations) {
+			glog.V(4).Infof("Deleting label %q for %q", constants.LabelBeforeReboot, n.Name)
+			glog.V(4).Infof("Setting annotation %q to true for %q", constants.AnnotationOkToReboot, n.Name)
+			err = k8sutil.UpdateNodeRetry(k.nc, n.Name, func(node *v1api.Node) {
+				delete(node.Labels, constants.LabelBeforeReboot)
+				// cleanup the before-reboot annotations
+				for _, annotation := range k.beforeRebootAnnotations {
+					delete(node.Annotations, annotation)
```
dghubble: Verbose log these deletions

pkg/operator/operator.go
sdemos: I'm just going to remove the if statement entirely so that it always logs how many rebooted nodes it found. That's the same thing that mark before reboot does too.

sdemos:
fixed usability nits and rebased onto master. @dghubble

PR #122, opened 2017-07-25
Add annotation to exclude nodes from being schedulable 

rajiteh:
This PR introduces a new annotation that cluster admins can add to specific nodes to prevent them from being "Schedulable" by the agent. This is specially useful when you have certain nodes that only run master components (ie: `node-role.kubernetes.io/master=true`) and want them to participate in the update orchestration.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -35,6 +35,12 @@ var (
 	}).AsSelector()
 )
 
+var (
+	shouldStayUnscheduledSelector = fields.Set(map[string]string{
+		constants.AnnotationSchedulingDisabled: constants.True,
+	}).AsSelector()
+)
+
 func New(node string) (*Klocksmith, error) {
 	// set up kubernetes in-cluster client
 	kc, err := k8sutil.GetClient("")
@@ -81,10 +87,18 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
 		return fmt.Errorf("failed to set node info: %v", err)
 	}
 
-	// we are schedulable now.
-	glog.Info("Marking node as schedulable")
-	if err := k8sutil.Unschedulable(k.nc, k.node, false); err != nil {
+	// we are schedulable now, but first ensure that this node can be schedulable
+	unschedulable, err := k8sutil.MatchNodeAnnotations(k.nc, k.node, shouldStayUnscheduledSelector)
+	if err != nil {
 		return err
+	}
+	if unschedulable {
+		glog.Info("Not altering node schedulability due to annotation")
+	} else {
+		glog.Info("Marking node as schedulable")
+		if err := k8sutil.Unschedulable(k.nc, k.node, false); err != nil {
+			return err
+		}
 	}
 
 	// set coreos.com/update1/reboot-in-progress=false and
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -51,6 +51,12 @@ const (
 	// It is an opaque string, but might be semver.
 	AnnotationNewVersion = Prefix + "new-version"
 
+	// Key set by the administrator to mark a node as default unschedulable.
+	// Never set by the update-agent or update-operator.
+	//
+	// Prevents update-agent from marking SchedulingDisabled nodes Schedulable.
+	AnnotationSchedulingDisabled = Prefix + "scheduling-disabled"
+
 	// Key set by the update-agent to the value of "ID" in /etc/os-release.
 	LabelID = Prefix + "id"
 
--- a/pkg/k8sutil/metadata.go
+++ b/pkg/k8sutil/metadata.go
@@ -80,6 +80,16 @@ func SetNodeAnnotations(nc v1core.NodeInterface, node string, m map[string]strin
 	})
 }
 
+// MatchNodeAnnotations matches a given selector to a node's
+// annotations.
+func MatchNodeAnnotations(nc v1core.NodeInterface, node string, selector fields.Selector) (bool, error) {
+	n, err := nc.Get(node, v1meta.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+	return selector.Matches(fields.Set(n.Annotations)), nil
+}
+
 // Unschedulable marks node as schedulable or unschedulable according to sched.
 func Unschedulable(nc v1core.NodeInterface, node string, sched bool) error {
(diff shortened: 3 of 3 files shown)


dghubble:
We already have a plan in place that allows users to control the `update-agent` daemonset and where it runs (e.g. node selector on `container-linux-update.v1.coreos.com/agent=true` is one option).

Adding a knob so an admin can add a label to have the agent still run, but not mark a node schedulable again doesn't seem to buy much you couldn't get by just editing the agent=true/false toggle. In both cases, the admin marking the node down for maintenance (or Unshcedulable for other reasons) has to add a special label. But here, the Container Linux version can still upgrade. I worry slightly about how this will play out with plans to health check that workloads can run post-reboot and adding complexity to the pre-reboot / post-reboot hooks design for custom health checker pods (#121). There's also the case that if this is set on nodes erroneously, no issues occur immediately but the next upgrade leaves all nodes unschedulable.

I'd like to know more about the use case to be convinced its worth the complexity cost or having another knob. Would having fine grained control over where update-agent runs be sufficient?

euank:
@dghubble I think you misunderstand the issue.

Right now, the update-agent, after rebooting a node, "un-cordons" it (marks it schedulable).

That's not desirable for nodes that previously were marked unschedulable.

The simple goal here is "I want to update my master nodes with CLUO, but not have them become schedulable".

I think the right solution isn't adding a new knob, but rather "remembering" whether a node was unschedulable previously and restoring it to that previous state after the update (whether that be uncordoning it in the default case, or leaving it unschedulable if it already was before the drain).

euank:
@rajiteh What do you think of adding an annotation of "node-was-unschedulable" that is set to true/false just before "drain", and then restoring the unschedulable state based on said annotation?

If that covers your use-case, I'd prefer that change to adding a new config option.

rajiteh:
> I think the right solution isn't adding a new knob, but rather "remembering" whether a node was unschedulable previously.

I thought about that approach too but decided it's better to be explicit than implicit. It also brings in more complexity, ie: a previously CLUO labelled Node gets cordoned by the admin.

rajiteh:
@euank wouldn't a node be unschedulable regardless after an update reboot? Then we need a first run check. Either way, it still makes it impossible to manually cordon a node after it was "claimed" by the agent.

dghubble:
@euank Yes, that's my understanding as well and aligns with the comments here. I have the same concerns and desire to hear the usage scenario.

If a node has been marked unscheduled by an administrator, that's typically to address a problem. When I'm in this scenario, it makes sense to remove update-agent from the node entirely while debugging the problem. In that scenario, there isn't a need for a special mode that allows the node to keep being part of coordinated upgrades (let alone do the "right" thing with Schedulability).

Now, that's just one usage scenario. Let's hear the justification for this. Because it has downsides, like I mentioned. This issue gets at the heart of whether a cluster upgrade encompases ensuring nodes are in a healthy schedulable state (#121 allows users to define "health") or whether a complete upgrade can occur even though maybe no nodes are successfully doing work (in theory).

euank:
@rajiteh 
What I'm suggesting is that the agent stores, as a new annotation (meant as an internal implementation detail, not ever to be touched), whether the agent moved the node from schedulable -> unschedulable during a drain operation.

This happens late enough that it will in general pick up admin changes. If the agent runs, then the admin cordons the node, and then a reboot operation occurs afterwards, the agent would still leave the node cordoned. It need not be all that racy simply by updating both unschedulable and the new annotation in a single atomic update of the node object (which would fail if the administrator had cordoned/uncordoned in the meanwhile).

I don't think it adds all that much complexity and I think that it ends up being significantly less surprising behavior than having to learn of a new unschedulable configuration when using the agent.


@dghubble Apologies for my hasty and incorrect reading of your comment. Whoops!

That being said, I think the historical use of "unscheduled" on master nodes by most kubernetes deployment tools has removed much of the meaning of that taint. At this point we basically have to treat that taint as not indicating a real problem, but rather as (master node || drain in progress).

Perhaps the proper solution would even be for us to acknowledge that taint is overused and simply use our own taint, the "container-linux-updating" taint, which the agent could then tolerate as well as any other important components.... however, since afaik there isn't a way to tolerate all taints, the fact that master components currently tolerate the scheduleable taint means we're kinda stuck with it.

Apologies for that last bit being kinda thinking-out-loud.

dghubble:
For the controllers / masters, CoreOS clusters run the kubelet with `--register-with-taints=node-role.kubernetes.io/master=:NoSchedule` btw.

rajiteh:
@dghubble My use case specifically involves the fact that we have dedicated "master" nodes, which only run the `api-server`, `scheduler`, `kube-proxy` and `controller-manager`. Our provisioning scripts create Pod manifests for these services at `/etc/kubernetes/manifests` in each "master" node and starts kubelets with the argument ` --enforce-node-allocatable=''`. This is a pattern described by the docs here: https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/

When CLUO agent is deployed to the cluster as  a DaemonSet, it still creates a pod in these nodes and end up un-cordoning them. While we want our "master" nodes to be managed by CLUO, we don't want to make them available for scheduling to the rest of the cluster.

rajiteh:
@euank Ah. Sorry I didn't understand your comment at first. I think your approach makes sense. It satisfies my requirements so I'm good with that.

@dghubble I see. `--register-with-taints=node-role.kubernetes.io/master=:NoSchedule` is something we haven't looked in to. Perhaps that will resolve this problem (at-least for us) without any changes(?) 

Again, thank you for your input on this.

dghubble:
CoreOS example clusters and Tectonic clusters use that `--register-with-taints` flag on controller kubelets and aren't seeing behavior where workloads start running on controllers after upgrade. That'd be a bug and we'd wanna solve it outright over requiring annotations/labels.

This sounds like we've made an undocumented assumption about how users are doing workload isolation, we should mention it in requirements, and maybe add a few notes about differences between tainting controllers and marking them unscheduled.

rajiteh:
@dghubble @euank I do think its still valid use case that CLUO shouldn't un-cordon Nodes that weren't cordoned by it. I'm closing this PR because, like suggested, there are better ways of achieving that rather than using annotations. Also, my problem specifically is solved by using the taints, thus no changes were necessary.  

Thank you for your time and input! Cheers!

Issue #19, opened 2016-10-21
Documentation: documents labels and lifecycle of update


philips:
Please take the google doc and put it into the README so people know how this all works. The primary things will be the labels and overall theory of operation.

dghubble:
Documented labels at https://github.com/coreos/container-linux-update-operator/blob/master/doc/labels-and-annotations.md

dghubble:
I can't unassign myself, but I'm not working on this.

PR #123, opened 2017-08-01
doc: Document CLUO node labels and annotations

dghubble:
rel #19

### DIFF
--- a/doc/labels-and-annotations.md
+++ b/doc/labels-and-annotations.md
@@ -1,0 +1,40 @@
+# Node Labels and Annotations
+
+The CLUO `update-operator` and `update-agent` manage a set of node labels and annotations to coordinate reboots among nodes receiving `update_engine` updates. CLUO label and annotation names are prefixed with "container-linux-update.v1.coreos.com/" to avoid conflicts.
+
+A few labels may be set directly by admins to customize behavior. These are called out below. Other CLUO labels and annotations reflect coordinated state changes and should **not** be directly modified.
+
+## Update Operator (Coordinator)
+
+**Labels**
+
+| name  | example    | setter | description |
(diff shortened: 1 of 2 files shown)


doc/labels-and-annotations.md
```diff
@@ -0,0 +1,40 @@
...
+A few labels may be set directly by admins to customize behavior. These are called out below. Other CLUO labels and annotations reflect coordinated state changes and should **not** be directly modified.
+
+## Update Operator (Coordinator)
+
+**Labels**
+
+| name  | example    | setter | description |
+|-------|------------|--------|---------------|
+| agent | true/false | admin, update-operator | When the `auto-label-container-linux` compatability mode is enabled (via flag), the `update-operator` sets agent true on Container Linux nodes. This is a convenient label that users may node selector upon, if desired. |
+| reboot-paused  | true/false | admin | May be set to true by an admin so the `update-operator` will not consider a node for coordinated reboot (note: Container Linux updates can still occur) |
```
sdemos: When I read this it seems to imply that the operator will just reboot the node if it asks to be rebooted. I think you should remove the word coordinated, since it won't really consider it at all. Also, it should be mentioned more explicitly that reboots can still occur separately of cluo, instead of just mentioning the possibility for Container Linux updates.

dghubble: May be set to true by an admin so the `update-operator` will ignore a node requesting a reboot-needed. Note that CLUO only coordinates reboots, `update_agent` still installs updates which are applied when a node reboots (e.g. powerloss).

sdemos: `s/update_agent/update_engine/` but yeah

dghubble: thanks

Issue #126, opened 2017-08-08
Steady state logs claim a label is being set when its not


dghubble:
In the v0.3.0 release, the log line when there's nothing to be done gives the impression its labeling a node.

```
I0808 21:47:29.436535       1 operator.go:517] Found 0 rebooted nodes, setting label "container-linux-update.v1.coreos.com/after-reboot" to "true"
```

We decided we should just trim this down to "Found X rebooted nodes", the loop a few lines below calls `mark` which logs information about nodes the operator is labeling.

PR #130, opened 2017-08-17
v0.3.1 release

dghubble:
Release notes (proposed):

Fix log message in steady-state that erroneously claimed labels are changing (#126)


VERSION
```diff
@@ -1 +1 @@
-0.4.0-dev
```
dgonyeo: Why are there two commits that both touch `VERSION`, with the second cancelling out the change the first one made to the file?

dghubble: Look at [build/bump-release](build/bump-release). I'm not a huge fan of this way of doing it either, but that's the current setup.

PR #132, opened 2017-09-05
update-agent: Use default grace period for DeleteOptions when draining node

hankjacobs:
By omitting a grace period, the object's default grace period will be used. This should allow pods with non-default termination grace periods to exit cleanly.


bkeroackdsc:
Ping. Is there any reason this can't be merged?

dghubble:
Seems reasonable to me.

PR #137, opened 2017-09-29
update-agent: wait for pods to terminate before rebooting

hankjacobs:
This change makes the agent wait until all pods are terminated before rebooting. Prior to this change, the agent would reboot immediately after the pod delete request was made and, consequently, pods were not given the time needed to terminate gracefully.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -7,6 +7,7 @@ import (
 	"github.com/coreos/go-systemd/login1"
 	"github.com/golang/glog"
 	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	v1meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -155,6 +156,15 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
 		if err := k.kc.CoreV1().Pods(pod.Namespace).Delete(pod.Name, deleteOptions); err != nil {
 			glog.Errorf("failed terminating pod %q: %v", pod.Name, err)
 			// Continue anyways, the reboot should terminate it
+		}
+	}
+
+	for _, pod := range pods {
+		glog.Infof("Waiting for pod %q to terminate", pod.Name)
+		if err := k.waitForPodDeletion(pod); err != nil {
+			glog.Errorf("failed waiting for pod %q to terminate: %v", pod.Name, err)
+		} else {
+			glog.Infof("Pod %q terminated", pod.Name)
 		}
 	}
 
@@ -345,6 +355,20 @@ func (k *Klocksmith) getPodsForDeletion() ([]v1.Pod, error) {
 	})
 
 	return pods, nil
+}
+
+// waitForPodDeletion waits for a pod to be deleted
+func (k *Klocksmith) waitForPodDeletion(pod v1.Pod) error {
+	return wait.PollInfinite(time.Second, func() (bool, error) {
+		switch _, err := k.kc.CoreV1().Pods(pod.Namespace).Get(pod.Name, v1meta.GetOptions{}); {
+		case err == nil:
+			return false, nil
+		case errors.IsNotFound(err):
+			return true, nil
+		default:
+			return false, err
+		}
+	})
 }
 
 // sleepOrDone pauses the current goroutine until the done channel receives
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -27,7 +27,11 @@ type Klocksmith struct {
 	nc   v1core.NodeInterface
 	ue   *updateengine.Client
 	lc   *login1.Conn
+	ri   time.Duration
+	rt   time.Duration
 }
+
+const reapInterval = 10 * time.Second
 
 var (
 	shouldRebootSelector = fields.Set(map[string]string{
@@ -36,7 +40,7 @@ var (
 	}).AsSelector()
 )
 
-func New(node string) (*Klocksmith, error) {
+func New(node string, reapTimeout time.Duration) (*Klocksmith, error) {
 	// set up kubernetes in-cluster client
 	kc, err := k8sutil.GetClient("")
 	if err != nil {
@@ -58,7 +62,7 @@ func New(node string) (*Klocksmith, error) {
 		return nil, fmt.Errorf("error establishing connection to logind dbus: %v", err)
 	}
 
-	return &Klocksmith{node, kc, nc, ue, lc}, nil
+	return &Klocksmith{node, kc, nc, ue, lc, reapInterval, reapTimeout}, nil
 }
 
 // Run starts the agent to listen for an update_engine reboot signal and react
@@ -359,7 +363,7 @@ func (k *Klocksmith) getPodsForDeletion() ([]v1.Pod, error) {
 
 // waitForPodDeletion waits for a pod to be deleted
 func (k *Klocksmith) waitForPodDeletion(pod v1.Pod) error {
-	return wait.PollInfinite(time.Second, func() (bool, error) {
+	return wait.PollImmediate(k.ri, k.rt, func() (bool, error) {
 		switch _, err := k.kc.CoreV1().Pods(pod.Namespace).Get(pod.Name, v1meta.GetOptions{}); {
 		case err == nil:
 			return false, nil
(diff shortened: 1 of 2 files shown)


hankjacobs:
Just a friendly ping on this

pkg/agent/agent.go
```diff
@@ -347,6 +357,20 @@ func (k *Klocksmith) getPodsForDeletion() ([]v1.Pod, error) {
...
 
+// waitForPodDeletion waits for a pod to be deleted
+func (k *Klocksmith) waitForPodDeletion(pod v1.Pod) error {
+	return wait.PollInfinite(time.Second, func() (bool, error) {
+		switch _, err := k.kc.CoreV1().Pods(pod.Namespace).Get(pod.Name, v1meta.GetOptions{}); {
+		case err == nil:
+			return false, nil
+		case errors.IsNotFound(err):
+			return true, nil
+		default:
```
dghubble: Handling this more naively could improve logs and retry. Compared with logging at the top level where the wait return error is just a loose approximation of what happened.

```
_, err := k.kc.CoreV1().Pods(pod.Namespace).Get(pod.Name, v1meta.GetOptions{})
if errors.IsNotFound(err) {
    glog.Infof("deleted pod %s", pod.Name)   // success log
	return true, nil
}
if err != nil {
  glog.Errorf("failed to get pod %v", err)    // some API call error, likely transient
}
return false, nil
```

```
Waiting for pod POD to terminate
failed to get pod: transient error
failed to get pod: transient error
deleted pod POD
```

or 

```
Waiting for pod POD to terminate
failed to get pod: transient error
failed to get pod: transient error
Skipping wait on pod POD
```

pkg/agent/agent.go
```diff
@@ -158,6 +159,15 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
 		}
 	}
 
+	for _, pod := range pods {
+		glog.Infof("Waiting for pod %q to terminate", pod.Name)
+		if err := k.waitForPodDeletion(pod); err != nil {
+			glog.Errorf("failed waiting for pod %q to terminate: %v", pod.Name, err)
```
dghubble: In practice, many of these log lines will be about transient errors to make API calls and we shouldn't stop the wait for them (e.g. "failed waiting for pod xx to terminate: can't reach 10.3.0.1"). Other uses of wait in agent allow the anon function to log errors and return nil to continue trying.

dghubble: On timeout, we can say we're skipping waiting for the pod to terminate.

pkg/agent/agent.go
```diff
@@ -347,6 +357,20 @@ func (k *Klocksmith) getPodsForDeletion() ([]v1.Pod, error) {
 	return pods, nil
 }
 
+// waitForPodDeletion waits for a pod to be deleted
+func (k *Klocksmith) waitForPodDeletion(pod v1.Pod) error {
+	return wait.PollInfinite(time.Second, func() (bool, error) {
```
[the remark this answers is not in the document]
hankjacobs: Any opposition to making the reap timeout configurable? 10 minutes seems reasonable for most use cases but there are some that require a much longer timeout. Specifically, some of our pods running in our cluster can take up to 24 hours to terminate gracefully in certain scenarios.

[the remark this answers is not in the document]
dghubble: I'm not greatly in favor of new flags, unless really needed. Apps that require 24 hours to terminate gracefully, on clusters designed around the idea that individual nodes can go down and workloads move flexibly?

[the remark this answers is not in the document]
dghubble: If you've been using the original PR on clusters (and we mentioned that on any transient call error, the wait would exit), it sounds like your apps are already subject to ungraceful shutdown?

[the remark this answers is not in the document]
hankjacobs: We haven't been using it on our clusters for the very reason this PR was opened. The fact that pods don't terminate gracefully before reboot showed up in our testing prior to rolling this out.

Yep, it's most definitely not ideal that these jobs can take so long (most of the time they don't) and it's something that is being worked on but we have to work around it for the time being. 

24 hours is most definitely an outlier but I'd argue that 11 minutes isn't. 15 minutes still sounds pretty reasonable to me depending on what's running in the cluster. The point I'm trying to make is that the reap timeout is heavily dependent on the type of workload that runs on your cluster. Because it could be so variable, I'd say it's worth making into a flag.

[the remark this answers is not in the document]
hankjacobs: Ah sorry, misunderstood that last question. We’re used this PR on our test cluster running a similar workload but with not nearly as long of running.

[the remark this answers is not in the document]
dghubble: Hm, I'm not excited to add config flags (flags cascade into higher-level components), but I can see a `--grace-period` flag on update-agent (mirroring the --grace-period flag of `kubectl drain`) being warranted here. It does also set expectations that there has to be some period of time beyond which we give up on waiting (i.e. its best-effort, we can't guarantee we'll satisfy every pod grace period).

hankjacobs:
@dghubble feedback implemented. thanks!

FaKod:
Just a comment on this. I am using Rook (a Ceph based cluster filesystem) and failed with locksmith and this PR to reboot updated nodes. 
This PR doesn't help because Rooks Pods (OSDs) are killed before the Pods of f.e. ElasticSearch are killed. So the ES Pods cannot be terminated at all by K8s and the agent waits forever.

hankjacobs:
@FaKod If graceful termination of a pod is dependent on the order in which other pods are terminated, this PR won't help much. That being said, the agent shouldn't hang with the changes I pushed ~6 hours ago. The agent should timeout after 10 minutes of waiting for a pod to terminate. Did you try with that version? If so, could you send me the agent logs? If not, could you try with that version? My testing proved successful in our test cluster.

FaKod:
@hankjacobs In case of rook: Although pods are terminated eventually, the mounts are still valid on the node. That prevents a node from restarting. See https://github.com/rook/rook/issues/1018

This may be a specific issue of Rook. What would help is to call "drain" before and wait for the non DaemonSet Pods to terminate.

hankjacobs:
This is effectively a "drain" as implemented by [kubectl](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/drain.go#L598) (albeit a simplified version). The hanging you saw was an issue with my original implementation. That should be fixed with the current version of this PR.

hankjacobs:
Additionally, [DS pods are skipped ](https://github.com/hankjacobs/container-linux-update-operator/blob/8f1b849444a0037f4a4d558696856ad578692f39/pkg/drain/drain.go#L21). To me, this points to something with your setup. This is outside the scope of this PR but I'd double check that you're running rook as a DS.

FaKod:
Right, its a ReplicaSet (created by an Operator) - sorry for the confusion. So I'll issue a feature request to exclude some Resources from deletion by labels or so.

pkg/agent/agent.go
```diff
@@ -26,16 +27,20 @@ type Klocksmith struct {
 	nc   v1core.NodeInterface
 	ue   *updateengine.Client
 	lc   *login1.Conn
+	ri   time.Duration
```
dghubble: remove, this isn't customized anywhere

pkg/agent/agent.go
```diff
@@ -26,16 +27,20 @@ type Klocksmith struct {
...
+	rt   time.Duration
```
dghubble: Can you name this reapTimeout or gracePeriod? No need to emulate these (poor imo) existing field names

pkg/agent/agent.go
```diff
@@ -26,16 +27,20 @@ type Klocksmith struct {
...
 }
 
+const reapInterval = 10 * time.Second
```
dghubble: Let's have this be defaultInterval or something more general to replace the other occurrences of 10 second intervals

pkg/agent/agent.go
```diff
@@ -57,7 +62,7 @@ func New(node string) (*Klocksmith, error) {
 		return nil, fmt.Errorf("error establishing connection to logind dbus: %v", err)
 	}
 
-	return &Klocksmith{node, kc, nc, ue, lc}, nil
+	return &Klocksmith{node, kc, nc, ue, lc, reapInterval, reapTimeout}, nil
```
dghubble: No need to configure with the constant

sdemos:
adding the `--grace-period` flag seems reasonable to me.

hankjacobs:
nits picked and an image to go with it: quay.io/dollarshaveclub/container-linux-update-operator:8015b109a973f0e8497ae894b90f66b6e47e5455-dirty

PR #140, opened 2017-10-06
Update kubernetes packages

dghubble:
* Update kubernetes, client-go, and apimachinery to versions used in KVO
* Switch to using client-go ConfigMap leader election instead of endpoints
* Requires client-go newer than v0.4.0 (latest tagged release) for leader election changes.
    * Downside to this is that its hard to say exactly how many Kubernetes versions in the future this is compatible with. On the other hand, getting all the kubernetes packages to match up properly is so chaotic, I'm not convinced "official" releases are that much more of a guarantee

Reasons:
* client-go v3.0.0-beta is quite old. We should update asap now that Kubernetes 1.8 is out.


dghubble:
Test image: `quay.io/dghubble/container-linux-update-operator:c38884b70061ee98ed2b5ab15e0c9da213c027ce-dirty` (dirty bc the build scripts assume sudoless docker).

squeed:
For my own knowledge, why do you need a pre-release version of client-go? I thought ConfigMap elections were in k8s 1.7!?!?

dghubble:
Leader election hadn't been moved to client-go yet.
* https://github.com/kubernetes/client-go/tree/v4.0.0/tools
* https://github.com/kubernetes/client-go/tree/c7ed6bc9c1c981e0f0bd09dc046c9b81ab855c24/tools

If you compare with KVO (internal), this glide.yaml should match what's used there to perform election in other operators.

squeed:
Cool, thanks for the info.

PR #141, opened 2017-10-10
get pods missing

FaKod:


FaKod:
oh sorry, I tested the "wait for pods termination" PR. So this PR is wrong

PR #143, opened 2017-10-16
Fix Jenkinsfile to test with Go 1.8.4

dghubble:
* Container images are built with Go 1.8.4 already

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,6 +22,6 @@ node('docker') {
     checkout scm
   }
   stage('Test') {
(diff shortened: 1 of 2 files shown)


sdemos:
as long as we are fixing go version numbers, a quick grep revealed that `build/build-release.sh` is still using `golang:1.8.3`

dghubble:
You're right. Ok so we have Jenkinsfile, build/build-image.sh, and build/build-release.sh.

I'd like to change this soon so Jenkins or Travis always produces a container image with the sha or tag (official releases).

PR #144, opened 2017-10-17
pkg/analytics: Remove Google Analytics tracking

dghubble:
We never even look at these analytics and its unclear what value they provide. No need to tradeoff being creepy if it isn't providing clear value.


cmd/update-operator/main.go
```diff
@@ -18,7 +17,6 @@ var (
 	beforeRebootAnnotations flagutil.StringSliceFlag
 	afterRebootAnnotations  flagutil.StringSliceFlag
 	kubeconfig              = flag.String("kubeconfig", "", "Path to a kubeconfig file. Default to the in-cluster config if not provided.")
-	analyticsEnabled        = flag.Bool("analytics", true, "Send analytics to Google Analytics")
```
euank: note, this will break people with `-analytics=false` set; it might be worth leaving the flag and printing a warning if it's set for a few releases.

dghubble: I was thinking a release note mentioning its removal in v0.4.0, but not opposed to leaving it

euank: Make sure to call it out as a breaking change. I'm okay either way if it's called out correctly.

Since we don't currently have a process for that though, it might be better to err on the side of giving longer notice.

dghubble: Sure

dghubble:
Added change to show a warning if `-analytics` flag is used (regardless of value). Rebased.

PR #148, opened 2017-10-23
pkg/updateengine: Return the right error on Call failures

dghubble:
On call errors, the `err` from the `Hello` was returned instead of the `call.Err` reason for the call failure.


sdemos:
actually, is there any reason why you didn't also remove the named returns on the `New` function? it looks like they are defined there but the behavior is ignored and everything is just directly returning instead.

dghubble:
Yeah, we could drop the named returns.

```
c := new(Client)
var err error
```

PR #150, opened 2017-10-24
agent: deserialize waitForPodDeletion check.

### LINKED ISSUE #149: CLUO uses pod name during deletion checks
The CLUO only checks the pod name for eviction checks. This is unsound, as pod names are not guaranteed to be unique or tied to a specific node.

https://github.com/coreos/container-linux-update-operator/blob/master/pkg/agent/agent.go#L362

It also serializes these checks for all pods where really they should be concurrent.

diegs:
This adds a sync.WaitGroup to check all pods concurrently and avoid
erroneously extended grace periods.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/coreos/go-systemd/login1"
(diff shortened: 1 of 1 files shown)


dghubble:
Let's have the WaitGroups be separate and do the UID to continue mirroring upstream drain as in #151.

diegs:
Ok, do you want me to scope this PR just to the waitgroup fix then?

dghubble:
Sure

diegs:
Rebased, though strictly not necessary since no conflicts... cc @euank :)

euank:
Our repo allows merge commits. Was it indicating otherwise @dghubble?

dghubble:
Habit, looks like its not enforced, which I'm fine with

PR #153, opened 2017-10-26
helm-chart: Create initial helm chart

hankjacobs:
This PR adds support for installing container-linux-update-operator via `helm`.


dghubble:
I'm really not interested in supporting Helm. We provide example manifests.

hankjacobs:
fair enough! I'll keep my branch around if you change your mind or if anyone else needs it.

PR #160, opened 2017-11-15
Jenkinsfile: Use declarative Jenkins pipeline

dghubble:
New multi-branch pipeline.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,26 +1,20 @@
-#!groovy
+pipeline {
+  agent none
 
-properties([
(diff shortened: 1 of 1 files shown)


euank:
This does mean PRs can't be "ok to test"ed and just will all be unilaterally un-tested-by-jenkins, right?

dghubble:
Yes, right now only PRs from "Contributors" will be run automatically. I'm not positive of the approval process so we may hold off on this.

I do still believe this will become the more sane direction to go in over time instead of ghprb.

dghubble:
I'm not really worried about changing this right now.

Issue #167, opened 2018-01-18
agent: set a label on nodes with a new OS


squeed:
Right now, the agent sets the `container-linux-update.v1.coreos.com/reboot-needed` annotation. However, I'd like the opportunity to run tectonic-torcx downloads ASAP, not just when the node has the reboot lock. This means we need a label too.

PR #168, opened 2018-01-18
 agent: set a label when after successful os-update 

squeed:
agent: set a label when after successful os-update
    
This sets the label `container-linux-update.v1.coreos.com/os-update-staged: "true"` on nodes where the OS has been upgraded but not yet rebooted. This is so daemonsets like tectonic-torcx can run ASAP.
    
Fixes: #167

### DIFF
--- a/doc/labels-and-annotations.md
+++ b/doc/labels-and-annotations.md
@@ -30,12 +30,13 @@ A few labels may be set directly by admins to customize behavior. These are call
 | id   | coreos  |  update-agent    | Reflects the ID in `/etc/os-release` |
 | version | 1497.7.0 | update-agent | Reflects the VERSION in `/etc/os-release` |
 | group | stable | update-agent     | Reflects the GROUP in `/usr/share/coreos/update.conf` or `/etc/coreos/update.conf` |
+| reboot-needed | true | update-agent | Reflects the reboot-needed annotation |
 
 **Annotations**
 
 | name | example | setter           | description |
 |------|---------|------------------|-------------|
-| reboot-needed  | true/false | update-agent | Set to true to request a coordinated reboot |
+| reboot-needed  | true/false | update-agent | Updates to true to request a coordinated reboot from the operator |
 | reboot-in-progress | true/false | update-agent | Set to true to indicate a reboot is in progress |
 | status | UPDATE_STATUS_IDLE | update-agent | Reflects the `update_engine` CurrentOperation status value |
 | new-version       | 0.0.0      | update-agent | Reflects the `update_engine` NewVersion status value |
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -92,10 +92,14 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
 		constants.AnnotationRebootInProgress: constants.False,
 		constants.AnnotationRebootNeeded:     constants.False,
 	}
+	labels := map[string]string{
+		constants.LabelRebootNeeded: constants.False,
+	}
 	glog.Infof("Setting annotations %#v", anno)
-	if err := k8sutil.SetNodeAnnotations(k.nc, k.node, anno); err != nil {
+	if err := k8sutil.SetNodeAnnotationsLabels(k.nc, k.node, anno, labels); err != nil {
 		return err
 	}
+
 	// Since we set 'reboot-needed=false', 'ok-to-reboot' should clear.
 	// Wait for it to do so, else we might start reboot-looping
 	if err := k.waitForNotOkToReboot(); err != nil {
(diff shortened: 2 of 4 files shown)


squeed:
cc @lucab

pkg/constants/constants.go
```diff
@@ -72,6 +72,11 @@ const (
 	// auto-label-container-linux=true
 	LabelUpdateAgentEnabled = Prefix + "agent"
 
+	// Label set to "true" on nodes where the update-agent has installed a new
+	// version, and a reboot is needed. Equivalent to the "reboot-needed"
+	// annotation, except that other agents sometimes set the annotation
```
dghubble: huh? "Equivalent except other agents sometimes set the annotation"?

pkg/agent/agent.go
```diff
@@ -199,14 +203,17 @@ func (k *Klocksmith) updateStatusCallback(s updateengine.Status) {
 		constants.AnnotationNewVersion:      s.NewVersion,
 	}
 
+	labels := map[string]string{}
+
 	// indicate we need a reboot
 	if s.CurrentOperation == updateengine.UpdateStatusUpdatedNeedReboot {
 		glog.Info("Indicating a reboot is needed")
 		anno[constants.AnnotationRebootNeeded] = constants.True
+		labels[constants.LabelUpdateStaged] = constants.True
```
dghubble: Assuming we do need this label, why not have it be "reboot-needed"? Its effectively acting the same as the existing reboot-needed annotation and gets set true/false in the same locations.

We have a bunch of labels and annotations documented in https://github.com/coreos/container-linux-update-operator/blob/master/doc/labels-and-annotations.md#update-agent so I'm not keen to introduce a new one. Its complicated enough to explain them. At least with reboot-needed, folks are already familiar and we can say both a label and an annotation are set in lockstep.

pkg/k8sutil/metadata.go
```diff
@@ -80,6 +80,20 @@ func SetNodeAnnotations(nc v1core.NodeInterface, node string, m map[string]strin
 	})
 }
 
+// SetNodeAnnotationsLabels sets all keys in a and l to their respective values in
```
dghubble: likewise -> respectively.

pkg/k8sutil/metadata.go
```diff
@@ -80,6 +80,20 @@ func SetNodeAnnotations(nc v1core.NodeInterface, node string, m map[string]strin
...
+// node's annotations and labels, likewise
+func SetNodeAnnotationsLabels(nc v1core.NodeInterface, node string, a, l map[string]string) error {
+	return UpdateNodeRetry(nc, node, func(n *v1api.Node) {
```
dghubble: Or just call the SetNodeLabels and SetNodeAnnotation methods that are already there.

dghubble:
Ah, missed #167 which has the background context.

pkg/agent/agent.go
squeed: The only reason I didn't do that is because the reboot-needed annotation is sometimes set by other operators. But I think you're right.

pkg/constants/constants.go
squeed: I was referencing the behavior of the kvo, but I doubt this is worth mentioning.

pkg/k8sutil/metadata.go
squeed: Is there any value in doing it in a single transaction?

dghubble:
So can we have the label use the same name? And we'll document that we set the reboot-needed annotation and label in lock-step. Label is easier to write deployments / daemonsets against. Annotation is easier to write programmatic code against.

The fact Tectonic sets CLUO's annotations is an anti-pattern I've tried to squash. CLUO doesn't make guarantees when external apps are setting its own clearly named attributes.

squeed:
Thanks for the feedback; updated. PTAL.

doc/labels-and-annotations.md
```diff
@@ -30,6 +30,7 @@ A few labels may be set directly by admins to customize behavior. These are call
 | id   | coreos  |  update-agent    | Reflects the ID in `/etc/os-release` |
 | version | 1497.7.0 | update-agent | Reflects the VERSION in `/etc/os-release` |
 | group | stable | update-agent     | Reflects the GROUP in `/usr/share/coreos/update.conf` or `/etc/coreos/update.conf` |
+| reboot-needed | true | update-agent | Set to true to request a coordinated reboot |
```
sdemos: this should explicitly say that this label is required to be in lockstep with the associated annotation.

dghubble: While we're at it, maybe we should say "Updates to true to request a coordinated reboot". I can see someone misreading "Set" as an instruction to manipulate it themselves.

dghubble:
I have no other issues. Just the point by @sdemos. We can do ship this feature out if needed or wait a bit after merge.

squeed:
Thanks for the feedback, updated. It would probably be good to release this soon.

PR #169, opened 2018-01-25
Release v0.6.0

dghubble:
Release notes:

* Add the `container-linux-update.v1.coreos.com/reboot-needed: "true"` label to reflect the "reboot-needed" annotation (#168)


sdemos:
I don't feel strongly either way but does this warrant a full minor version release? especially so soon after the previous one. if you have a particular reason for it though, it seems fine to me.

dghubble:
Yeah, I'm not so thrilled to have such a tiny release either. Its slightly more of a feature than a fix was my logic, but I could see either way.

sdemos:
I guess since it does things like break the kvo when it tries to mess with our annotations, a minor version is probably warranted. let's just leave it as is.

dghubble:
Yeah. And we'd wait for more features, but Tectonic seems to need this one.

PR #170, opened 2018-01-30
Feature/reboot window

mkoerperriag:
Implemented support for configuring a reboot window, using timeutil/periodic from locksmith.
Would be glad if you would review/merge this.

### DIFF
--- a/doc/reboot-windows.md
+++ b/doc/reboot-windows.md
@@ -1,0 +1,37 @@
+# Reboot windows
+The CLUO `update-operator` can be configured to only reboot nodes during certain timeframes.
+Pre-reboot checks are prevented from running as well.
+
+## Configuring update-operator
+
+The reboot window is configured through the the flags `--reboot-window-start` 
+and `--reboot-window-length`, or through the environment variables 
+`UPDATE_OPERATOR_REBOOT_WINDOW_START` and `UPDATE_OPERATOR_REBOOT_WINDOW_LENGTH`.
+
+Here is an example configuration:
+
+```
+/bin/update-operator \
+ --reboot-window-start=14:00 \
+ --reboot-window-length=1h
+```
+
+This would configure `update-operator` to only reboot between 2pm and 3pm. Optionally,
+a day of week may be specified for the start of the window:
+
+```
+/bin/update-operator \
+ --reboot-window-start="Thu 23:00" \
+ --reboot-window-length=1h30m
+```
+
+This would configure `update-operator` to only reboot the system on Thursday after 11pm,
+or on Friday before 12:30am.
+
+Currently, the only supported values for the day of week are short day names,
+e.g. `Sun`, `Mon`, `Tue`, `Wed`, `Thu`, `Fri`, and `Sat`, but the day of week can
+be upper or lower case. The time of day must be specified in 24-hour time format.
+The window length is expressed as input to go's [time.ParseDuration][time.ParseDuration]
+function.
+
+[time.ParseDuration]: http://godoc.org/time#ParseDuration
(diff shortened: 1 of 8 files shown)


sdemos:
Thanks for the pr! Things have been a little busy this week, so I haven't had the time to take a look at this as much as I would like to. 

I do have one quick question though before I look closer at it. Why did you decide to bump the waiting time period in the agent? Logging an error every 24 hours is a little messy, but it just loops around and watches again anyway. It already has the behavior of restarting this watching period while it's waiting for update_engine to request a reboot anyway, so I don't see the need for an unrelated change here, especially since the periodic library already doesn't allow periods longer than 24 hours. 

Also, sorry the jenkins build failed before, it was acting up earlier this week, I reran the job manually and it passed.

mkoerperriag:
I decided to bump it, because the reboot window may be on a weekly basis.
If you think its unnecessary, I'm completely comfortable with it.

sdemos:
yeah, I think we should just leave it the way it is for now, it needs fixing up in other ways anyway. 

Everything else in this pr seems okay to me. @dghubble do you have any comments on it?

dghubble:
No explicit problems. I'm rather unhappy to add features given the state of the repo and lack of tests.

dghubble:
Would be nicer to add docs akin to those locksmith provides. https://github.com/coreos/locksmith#reboot-windows

debtcollapse:
Can one of the admins verify this patch?

debtcollapse:
Can one of the admins verify this patch?

pablokbs:
@mkoerperriag Hello, I'm trying your changes but I can't make it work, should I add the window annotations to the nodes with the agent? As annotations?

```
kubectl annotate nodes ip-10-80-25-82.ec2.internal "container-linux-update.v1.coreos.com/reboot-window-start=Mon 22:00" --overwrite
kubectl annotate nodes ip-10-80-25-82.ec2.internal "container-linux-update.v1.coreos.com/reboot-window-lenght=10m" --overwrite
```

sdemos:
@pablokbs based on the implementation, the configuration is done through command-line flags, not annotations, so you have to modify your operator deployment with the new flags to take advantage of this behavior. Environment variables of the form `UPDATE_OPERATOR_REBOOT_WINDOW_START` and `UPDATE_OPERATOR_REBOOT_WINDOW_LENGTH` should also work, also set in the operator deployment.

@mkoerperriag sorry, I dropped the ball a little bit on this one. Can you add the documentation that @dghubble mentioned in his comments describing how the reboot windows work? I would place them under the `doc/` folder in their own markdown file.

mkoerperriag:
Thanks for your feedback so far.
@sdemos I will add the docs in the next couple of days.

doc/reboot-windows.md
```diff
@@ -0,0 +1,35 @@
+# Reboot windows
+The CLUO `update-operator` can be configured to only reboot nodes during certain timeframes.
+
+## Configuring update-operator
+
+The reboot window is configured through the the flags `--reboot-window-start` 
+and `--reboot-window-length`.
```
sdemos: we should probably mention the environment variables that can be set to configure these as well (they come for free from the flag definition, see https://github.com/coreos/container-linux-update-operator/blob/master/cmd/update-operator/main.go#L37)

doc/reboot-windows.md
```diff
@@ -0,0 +1,35 @@
+# Reboot windows
+The CLUO `update-operator` can be configured to only reboot nodes during certain timeframes.
```
sdemos: add something like "Pre-reboot checks are prevented from running as well." since we check the window before applying the before-reboot label.

PR #176, opened 2018-02-23
Only Mark Node Schedulable If Agent Marked Unschedulable 

hankjacobs:
This PR changes the agent to mark a node schedulable only if the agent was responsible for marking it unschedulable.

Prior to this change, the agent would always mark a node as schedulable when starting up. This could result in a node that had been marked unschedulable by an external process (eg. by an admin using kubectl) being marked schedulable by it's update agent if the agent restarted for some reason.

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -123,7 +123,16 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
 			return err
 		}
 	} else {
-		glog.Info("Skipping marking node as schedulable -- node was unschedulable prior to update reboot")
+		glog.Info("Skipping marking node as schedulable -- node was marked unschedulable by an external source")
+	}
+
+	anno = map[string]string{
+		constants.AnnotationAgentMadeUnschedulable: constants.False,
+	}
+
+	glog.Infof("Setting annotations %#v", anno)
+	if err := k8sutil.SetNodeAnnotations(k.nc, k.node, anno); err != nil {
+		return err
 	}
 
 	// watch update engine for status updates
@@ -140,22 +149,26 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
 		glog.Warningf("error waiting for an ok-to-reboot: %v", err)
 	}
 
-	// set constants.AnnotationRebootInProgress and drain self
-	anno = map[string]string{
-		constants.AnnotationRebootInProgress: constants.True,
-	}
-
-	glog.Infof("Setting annotations %#v", anno)
-	if err := k8sutil.SetNodeAnnotations(k.nc, k.node, anno); err != nil {
-		return err
-	}
-
 	glog.Info("Checking if node is already unschedulable")
 	node, err = k8sutil.GetNodeRetry(k.nc, k.node)
 	if err != nil {
 		return err
 	}
 	alreadyUnschedulable := node.Spec.Unschedulable
+
+	// set constants.AnnotationRebootInProgress and drain self
+	anno = map[string]string{
+		constants.AnnotationRebootInProgress: constants.True,
+	}
+
+	if !alreadyUnschedulable {
+		anno[constants.AnnotationAgentMadeUnschedulable] = constants.True
+	}
+
+	glog.Infof("Setting annotations %#v", anno)
+	if err := k8sutil.SetNodeAnnotations(k.nc, k.node, anno); err != nil {
+		return err
+	}
 
 	// drain self equates to:
 	// 1. set Unschedulable if necessary
@@ -164,18 +177,9 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
 	// ('any pods that are neither mirror pods nor managed by
 	// ReplicationController, ReplicaSet, DaemonSet or Job')
 
-	if alreadyUnschedulable == false {
+	if !alreadyUnschedulable {
 		glog.Info("Marking node as unschedulable")
 		if err := k8sutil.Unschedulable(k.nc, k.node, true); err != nil {
-			return err
-		}
-
-		// set constants.AnnotationRebootInProgress and drain self
-		anno = map[string]string{
-			constants.AnnotationAgentMadeUnschedulable: constants.True,
-		}
-
-		if err := k8sutil.SetNodeAnnotations(k.nc, k.node, anno); err != nil {
 			return err
 		}
 	} else {
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -86,6 +86,16 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
 		return fmt.Errorf("failed to set node info: %v", err)
 	}
 
+	glog.Info("Checking annotations")
+	node, err := k8sutil.GetNodeRetry(k.nc, k.node)
+	if err != nil {
+		return err
+	}
+
+	// Only make a node schedulable if a reboot was in progress. This prevents a node from being made schedulable
+	// if it was made unschedulable by something other thin the agent
+	makeSchedulable := node.Annotations[constants.AnnotationAgentMadeUnschedulable] == constants.True
+
 	// set coreos.com/update1/reboot-in-progress=false and
 	// coreos.com/update1/reboot-needed=false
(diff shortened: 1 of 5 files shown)


debtcollapse:
Can one of the admins verify this patch?

debtcollapse:
Can one of the admins verify this patch?

debtcollapse:
Can one of the admins verify this patch?

euank:
Instance of https://github.com/jenkinsci/ghprb-plugin/issues/292.

Honestly, we should probably move CoreOS projects related to k8s over to [prow](https://github.com/kubernetes/test-infra/tree/master/prow#prow) now that they've finished moving upstream off jenkins.

pkg/agent/agent.go
```diff
@@ -136,16 +151,34 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
...
 	// unlike `kubectl drain`, we do not care about emptyDir or orphan pods
 	// ('any pods that are neither mirror pods nor managed by
 	// ReplicationController, ReplicaSet, DaemonSet or Job')
 
-	glog.Info("Marking node as unschedulable")
-	if err := k8sutil.Unschedulable(k.nc, k.node, true); err != nil {
-		return err
+	if !alreadyUnschedulable {
+		glog.Info("Marking node as unschedulable")
+		if err := k8sutil.Unschedulable(k.nc, k.node, true); err != nil {
```
euank: I believe there's a slight race here.

If the node is marked unscheduleable, but the annotation call fails (e.g. due to apiserver issues or this code failing), the failure condition is that the node will be unschedulable and never be made schedulable again.

I think that the `k8sutil.SetNodeAnnotations` call several lines above should include this annotation if necessary.

I think that setting the annotation before marking it unschedulable leaves it in a state where it will ultimately do the right thing once it succeeds.

pkg/agent/agent.go
```diff
@@ -106,10 +117,14 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
 		return err
 	}
 
-	// we are schedulable now.
-	glog.Info("Marking node as schedulable")
-	if err := k8sutil.Unschedulable(k.nc, k.node, false); err != nil {
-		return err
+	if makeSchedulable {
```
euank: I think this should be moved to before the `SetNodeAnnotationLabels` for `MadeUnschedulable: False`. As is, I believe if it crashes or has issues with this call, it will not restore state correctly afterwards.

pkg/agent/agent.go
hankjacobs: Ah, right. Good catch. Thanks!

pkg/agent/agent.go
hankjacobs: I see your point but I'm concerned with moving it above `if err := k.waitForNotOkToReboot(); err != nil {`. From the comment above that line, it seems like there might be trouble if we preemptively mark the node as schedulable before waiting for ok-to-reboot to clear. How about if I set the `MadeUnschedulable` annotation to false after the `if makeSchedulable` block?

euank: Ah, inded, you're right. Moving the annotation down seems like it should fix the concern I have correctly.

sdemos:
In addition to @euank's comments, if you don't mind squashing your cleanup and merge commits together, that would be awesome. Probably just the first three commits you have should cover it pretty well.

Just ping us whenever that is done and we will give it a final review.

hankjacobs:
Will do! A bit swamped but have this on my to-do list.

sdemos:
@hankjacobs any updates on this? just a friendly ping, no rush.

hankjacobs:
@euank @sdemos Good to go. Thanks for your patience!

sdemos:
Thanks for pushing the fix! I'll take a look at this again sometime soon. 

My mistake with the issue reference, I had the link in my copy buffer and accidentally put it in a comment on an unrelated issue.

pkg/k8sutil/metadata.go
```diff
@@ -35,6 +35,22 @@ func NodeAnnotationCondition(selector fields.Selector) watch.ConditionFunc {
...
+// GetNodeRetry gets a node object, retrying up to DefaultBackoff number of times if it fails
+func GetNodeRetry(nc v1core.NodeInterface, node string) (*v1api.Node, error) {
+	var apiNode *v1api.Node
+	err := RetryOnError(DefaultBackoff, func() error {
+		n, getErr := nc.Get(node, v1meta.GetOptions{})
+		if getErr != nil {
+			return fmt.Errorf("failed to get node %q: %v", node, getErr)
+		}
+
+		apiNode = n
```
ashcrow: nit: Seems like ``apiNode`` could be set where ``n`` is set. Not worth blocking over though.

pkg/agent/agent.go
ashcrow: Looks like this was done at line [165](https://github.com/coreos/container-linux-update-operator/pull/176/files#diff-64c23753751bef052315a093bb62faeaR165).

pkg/k8sutil/metadata.go
jlebon: Hmm, OTOH doing it this way makes it easy to verify that `apiNode` can only ever be set if we're done retrying.

pkg/agent/agent.go
sdemos: I believe this got updated after this comment and github just didn't mark it as resolved

pkg/agent/agent.go
```diff
@@ -86,6 +86,16 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
...
 	}
 
+	glog.Info("Checking annotations")
+	node, err := k8sutil.GetNodeRetry(k.nc, k.node)
+	if err != nil {
+		return err
+	}
+
+	// Only make a node schedulable if a reboot was in progress. This prevents a node from being made schedulable
+	// if it was made unschedulable by something other thin the agent
```
jlebon: Minor/optional: `s/thin/than/`

pkg/agent/agent.go
```diff
@@ -123,7 +123,16 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
...
-		glog.Info("Skipping marking node as schedulable -- node was unschedulable prior to update reboot")
+		glog.Info("Skipping marking node as schedulable -- node was marked unschedulable by an external source")
+	}
+
+	anno = map[string]string{
+		constants.AnnotationAgentMadeUnschedulable: constants.False,
+	}
+
+	glog.Infof("Setting annotations %#v", anno)
+	if err := k8sutil.SetNodeAnnotations(k.nc, k.node, anno); err != nil {
```
jlebon: We only need to do this if we're the ones who marked the node unschedulable, right? I.e. we can move this hunk to under the `if makeSchedulable {` above.

pkg/agent/agent.go
```diff
@@ -86,6 +86,16 @@ func (k *Klocksmith) process(stop <-chan struct{}) error {
...
 
+	glog.Info("Checking annotations")
+	node, err := k8sutil.GetNodeRetry(k.nc, k.node)
+	if err != nil {
+		return err
+	}
+
+	// Only make a node schedulable if a reboot was in progress. This prevents a node from being made schedulable
+	// if it was made unschedulable by something other thin the agent
+	makeSchedulable := node.Annotations[constants.AnnotationAgentMadeUnschedulable] == constants.True
```
sdemos: When the agent runs for the first time, this annotation doesn't exist, so this value is set as false, and the log message that says "node was marked unschedulable by external source" is printed out, which could be confusing. We should probably use the map multi-return value to check for the key existence and use that in the if statement below.

pkg/agent/agent.go
sdemos: It seems like it can be moved. I agree with this.

hankjacobs:
@sdemos All set with the requested changes. I haven't had a chance to test these changes since I had tore down my original testing infrastructure a while back. I may have some time to set it up tomorrow but no guarantees. Would you be able to test? If not, no biggie.

sdemos:
Changes look good to me. I can test it out. Thanks for the quick fixes!

PR #182, opened 2018-06-19
pkg/agent: wait for node to become ready

crawford:
The agent needs to wait for the node to become ready before removing the
reboot-in-progress annotation. This is to ensure that the cluster
doesn't simultaneously lose more than the expected number of nodes.


crawford:
Actually, this might cause a problem for post-reboot hooks. I'm going to move this logic to the operator instead.

PR #183, opened 2018-06-19
pkg/operator: wait for nodes to become ready

crawford:
The operator needs to wait for the node to become ready before removing
the reboot-ok label. This is to better ensure that the operator doesn't
simultaneously lose more than the expected number of nodes in the event
that individual nodes take a long time to become ready.


crawford:
This has not yet been tested. Do not merge.

crawford:
We were never able to verify that this was the root cause of the failure that the customer was seeing. Closing.

Issue #191, opened 2019-03-28
Persistent volumes in stuck state after reboot


ghost:
It appears there is a race condition when using persistent volumes, where the pod is deleted and the node is rebooted, but the attached volume is still in the process of detaching. Once this happens, the persistent volume is stuck and must be manually removed and recreated.

I'm seeing this on AWS with EBS volumes.

metalmatze:
I'm not sure hat that relates to CLUO? Are you running its Pods with volumes?

ghost:
No, not running CLUO with volumes. Is it possible that when the operator terminates the pods, it reboots before PV are properly detached?

embik:
Hey @dustinmm80, are you by chance using the CSI implemention of EBS volumes?

I see something similar with the Cinder CSI driver and I suspect it to be related to `VolumeAttachment` resources or rather the fact that CSI components might not be fast enough to detach volumes before CLUO reboots the machine.

Just wanted to check in before investigating this.

embik:
Thinking about it, I'm not sure it's only related to CSI. But it's probably a major issue with StatefulSets, because Kubernetes won't create a new `statefulset-example-0` on another node before the old one finished deletion.

And only _after_ scheduling the new StatefulSet pod to a new node the CSI components will start churning and update the `VolumeAttachment`, which will unmount the volume on the old node and mount it on the new node. But that process takes a few second and the old node is already being rebooted.

yannh:
We had the same issue a few weeks ago - it turned out to be an issue with newer, Nitro instances (c5, t3, ...) - AWS investigated and claims to have now solved the problem. Were you seeing that problem with nitro instances? Are you still seeing the problem?

PR #192, opened 2019-06-19
update-agent: Added reboot-wait parameter

johannwagner:
This adds an reboot-wait parameter, which waits, after the last pod was terminated, an
fixed amount of time to finalize operations before reboot. This solves some problems
this storage provisioners like rook.

/cc: @martin31821

### DIFF
--- a/cmd/update-agent/main.go
+++ b/cmd/update-agent/main.go
@@ -17,6 +17,7 @@ var (
 	node         = flag.String("node", "", "Kubernetes node name")
 	printVersion = flag.Bool("version", false, "Print version and exit")
 	reapTimeout  = flag.Int("grace-period", 600, "Period of time in seconds given to a pod to terminate when rebooting for an update")
+	rebootWait   = flag.Int("reboot-wait", 0, "Period of time in seconds waiting after last pod deletion for reboot")
 )
 
 func main() {
@@ -37,7 +38,10 @@ func main() {
 	}
 
 	rt := time.Duration(*reapTimeout) * time.Second
-	a, err := agent.New(*node, rt)
+	rw := time.Duration(*rebootWait) * time.Second
+
+	glog.Infof("Waiting %v for reboot", rw)
+	a, err := agent.New(*node, rt, rw)
 	if err != nil {
 		glog.Fatalf("Failed to initialize %s: %v", os.Args[0], err)
 	}
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -29,6 +29,7 @@ type Klocksmith struct {
 	ue          *updateengine.Client
 	lc          *login1.Conn
 	reapTimeout time.Duration
+	rebootWait  time.Duration
 }
 
 const defaultPollInterval = 10 * time.Second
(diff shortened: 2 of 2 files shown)


lucab:
@johannwagner thanks for the PR! However this lacks quite a bit of accompanying context, and on the surface it seems to just be trying to hide an existing race.

What is the problem this is trying to solve related to rook? How do you determine what is a good reboot-wait time (and is there some upper-bound guarantee on the value)? Why isn't pod draining sufficient here?

embik:
Hi @lucab, not the author but hope I can provide some context as well.

This seems to be useful because of situations like #191 - beyond pod termination there are some downstream operations (e.g., unmounting a volume because it is going to be attached to another node - this can take a short period of time; I suppose that's why it's useful for rook) that are not fully grasped by CLUO at the moment.

Implementing the `reboot-wait` parameter would be a bit hacky, but it's a nice way around implementing checks for all possible downstream operations (which even might be proprietary).

johannwagner:
Jeep, that’s the reason. We want to be able to shutdown our CEPH Cluster properly, which only works, if everything shutdowns properly after the last Pod was terminated. @embik assumes the correct context.

dghubble:
If a drain isn't sufficient for your situation, CLUO supports before-reboot (and after-reboot) required annotations. A custom DaemonSet can run before/after a reboot to perform custom cleanup/setup actions (e.g. waiting a fixed period, implement Ceph shutdown, etc.) and then add the required annotation, which CLUO awaits.

https://github.com/coreos/container-linux-update-operator/blob/master/doc/before-after-reboot-checks.md

martin31821:
@dghubble We are already using custom wait parameters to cleanly unmount and reboot nodes, but at least the rbd driver does some internal network related tasks which we can't wait for at the moment.

cgwalters:
I think this would probably be better done as a systemd unit with an `ExecStop=` that would run at shutdown time and call some API that waits for whatever Rook/Ceph needs to do.

martin31821:
If you follow these instructions https://github.com/rook/rook/blob/master/Documentation/container-linux.md on a CoreOS cluster, it will cause the nodes to hang during shutdown because the network is already brought down by systemd before the rbd module is able to complete its unmount and unlocking the objects/drive, that's why our initial implementation used a delay of approx. 30s to ensure this will be finished. 

Basically the before-reboot hook runs before the update operator drains the node and maybe it's good to have the same hook mechanic being executed `after-drain-before-reboot`. 

WDYT?

embik:
@dghubble in that case the custom DaemonSet would be required to implement draining the node, because in the scenario described we need to `drain -> wait -> reboot`, not `wait -> drain -> reboot`. Which means it will reimplement logic already present in CLUO. The workflow would be `drain (custom logic) -> wait (custom logic) -> already cordoned, no drain (CLUO) -> reboot (CLUO)`.

Honestly, if I need to write custom logic and deploy a custom DaemonSet and all that, I'd probably just fork CLUO and apply this patch because it solves the problem well enough. I could do that, no problem, I just think you should be aware this patch could be helpful for production users.

Issue #196, opened 2019-11-13
CLUO doesn't work on k8s 1.16


betermieux:
Since DaemonSets are moved from 'extensions/v1beta1' to 'apps/v1', the agent doesn't exclude itself from node draining anymore and kills itself before calling reboot. This results in a continuously draining and repopulating node. The culprit seems to be:
https://github.com/coreos/container-linux-update-operator/blob/4bb1486f482bc9c365c71e126129e806b5a0fc97/pkg/drain/drain.go#L74
If I find some time, I will send a pull-request.

sdlarsen:
As a work-around (while waiting for #197), it's possible to enable serving of the deprecated resource names by adding this to the api-server: `--runtime-config=extensions/v1beta1/daemonsets=true`

mleklund:
Having this issue as well.

PR #197, opened 2019-11-14
K8s 1.16 compatibility

betermieux:
A first take on making CLUO ready for k8s > 1.16
- Changed 'ExtensionsV1beta1().DaemonSets' to 'AppsV1().DaemonSets'
- Updated packages 'k8s.io/apimachinery', 'k8s.io/client-go' and 'k8s.io/kubernetes' to stable 1.16.2 versions
- Needed to update some more dependend packages (also to current stable versions)
- There were major changes to the watch interface (moved from apimachinery to client-go). I tried my best to adapt the source files, but I am unsure about: https://github.com/betermieux/container-linux-update-operator/blob/2f5fe31992300915154c81ffcadbb731eae4d45c/pkg/operator/operator.go#L266
I have tested theses changes on a 1.16.2 cluster and everything seems to be working like before.  Some more tests regarding leaderelection seem to be necessary.


betermieux:
#196

dghubble:
Yep, with Kubernetes v1.16+ the CLUO agent drain loops, since the API change means it doesn't exclude itself as it did previously. This update makes sense, its in line with the last time drain logic had to update for Kubernetes v1.9. Looks good to me.

However, myself and the other folks who used to approve and release aren't with CoreOS now. Perhaps if this was preceded with a bump to the Go version to fix the Travis setup, a Red Hatter could approve.

